PR #188, opened 2014-05-03
Python 3.4 support. Fixed README.md

eirnym:


eirnym:
@amoffat Thank you for answer!

PR #213, opened 2014-12-31
fix args with context to clean up after an exception is thrown

jbeder:
As a related question, it looks like with contexts are deprecated, since they're not threadsafe. What would be a replacement for pushd?

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -2138,8 +2138,10 @@ def args(*args, **kwargs):
         key = key.lstrip("_")
         call_args[key] = value
 
-    yield
-    call_args.update(old_args)
+    try:
+        yield
+    finally:
+        call_args.update(old_args)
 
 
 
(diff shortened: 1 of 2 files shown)


jbeder:
Yeah, something like that sounds reasonable. To be honest, I didn't even know you could specify _cwd= on a command, and that's easy enough for my use case - I think the documentation (http://amoffat.github.io/sh/index.html) is a bit out of date :)

Also, naming the command sh.pushd is a bit misleading, since pushd is a normal bash command - so (I think) you'd expect to write

    sh.pushd('/tmp')
    # stuff
    sh.popd()

but sh.popd() doesn't exist (command can't be found, since it's a builtin), and of course, sh.pushd does something a bit different.

So I think something like a context is pretty reasonable.

amoffat:
Merged into `release-1.2` branch and will go out with it.  Thanks for your contribution

PR #222, opened 2015-01-20
Ignore compiled pyc files

jawshooah:
Very helpful for those of us using this as a submodule dependency. Keeps our `git status` clean :smiley:


jawshooah:
Ah, just saw all the other rejected PRs. Sorry about that. Closing now.

Issue #224, opened 2015-01-28
python2 _iter unicode issue


basaundi:
Having a file with unicode characters longer than 1024 bytes won't iterate over lines.

Example file `test.utf8.txt`:
```
1297    jpn     きみにちょっとしたものをもってきたよ。
4702    jpn     何かしてみましょう。
4703    jpn     私は眠らなければなりません。
4704    jpn     何してるの？
4705    jpn     今日は６月１８日で、ムーリエルの誕生日です！
4706    jpn     お誕生日おめでとうムーリエル！
4707    jpn     ムーリエルは２０歳になりました。
4708    jpn     パスワードは「Muiriel」です。
4709    jpn     すぐに戻ります。
4710    jpn     知らない。
4711    jpn     何と言ったら良いか分かりません。
4712    jpn     きりがない。
4713    jpn     何と言ったらいいか・・・。
4714    jpn     こいつは悪いウサギだった。
4715    jpn     私は山にいました。
4716    jpn     それは最近の写真？
4717    jpn     時間があるか分かりません。
4718    jpn     先程、どういうわけかマイクが入りませんでした。
4720    jpn     この世界の教育にはがっかりしてしまう。
4726    jpn     私は一日に100ユーロ稼ぎます。
4727    jpn     すぐに諦めて昼寝をするかも知れない。
4728    jpn     それはあなたが一人になりたくないからです。
4729    jpn     そんなことは起きないでしょう。
4730    jpn     彼は時々変です。
4731    jpn     あなたの勉強を邪魔しないようにします。
4732    jpn     みんなもそうなのかな、と思うことくらいしかできない。
4733    jpn     長い目で見れば違ってくると思います。
4734    jpn     心配しないで。
4735    jpn     会えなくて淋しい。
4736    jpn     明日帰ったら電話します。
4737    jpn     私はいつも不思議な性格の方が好きだった。
4738    jpn     眠った方がいいよ。
4739    jpn     行くよ。
4740    jpn     私は彼らにチケットをもう一枚送ってくれるように頼んだ。
4741    jpn     君は僕にとてもいらいらしている。
4742    jpn     私はそんな風には生きられない。
4743    jpn     私は一時期、天体物理学者になりたかった。
```

The following won't work in python2:
```python
from sh import cat
for line in cat('test.utf8.txt', _iter=True):
    print len(line)
```
I get that the contents of the file are two lines of 1024 and the rest.
```
1024
1024
163
```

I suspect somewhere it should read:
```python
[...]

PR #225, opened 2015-01-28
test for issue https://github.com/amoffat/sh/issues/224

basaundi:
so I went ahead and made a test for issue #224


amoffat:
thanks for the starting point, Ander

PR #229, opened 2015-02-11
Ignore all python-compiled files

arthurdarcet:
If i am not mistaken, python 3 generate its `pyc` files in a `__pycache__` directory.
The `.pyo` files are similar to `pyc` (with optimisation).
I haven't included `pyd` files since they are windows-only.


amoffat:
:+1:

PR #232, opened 2015-02-26
Fix never-ending loop of checking output of dead process

kabakchey:
I've met problem with original sh. Sometimes it doesn't exit when child process is dead.
My research brought me to this solution.
Please, take a look.

Comments on my patch are very appreciatied.


kabakchey:
Still problems.

amoffat:
what's your case for reproducing the problem?

kabakchey:
Hi there,

I've added answer as comment to pull request.

Regards
Dmytro Kabakchey

I
26.02.2015 18:16, Andrew Moffat пишет:
> s your case for reproducing the problem?

PR #233, opened 2015-02-26
Fix infinite loop of select.select when process is dead

kabakchey:
I met some problem with original sh. It were falling into infinite loop when process is dead, but sh was in the middle of stdout reading.
My research lead me to this solution. Please, take a look.

Comments are very appreciatied.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -153,19 +153,19 @@ class ErrorReturnCode(Exception):
     derived classes with the format: ErrorReturnCode_NNN where NNN is the exit
     code number.  the reason for this is it reduces boiler plate code when
     testing error return codes:
-    
+
         try:
             some_cmd()
         except ErrorReturnCode_12:
             print("couldn't do X")
-            
+
     vs:
         try:
             some_cmd()
         except ErrorReturnCode as e:
             if e.exit_code == 12:
                 print("couldn't do X")
-    
+
     it's not much of a savings, but i believe it makes the code easier to read """
 
     truncate_cap = 750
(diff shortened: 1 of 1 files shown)


kabakchey:
Further investigation brings me to 'pg_ctl' binary's poor implementation of daemonization. 'pg_ctl' binary is an utility for starting postgresql RDBMS as daemon. 'pg_ctl' executes 'postgres' binary as a child process and daemon. The stdout descriptor is passed from 'pg_ctl' to 'postgres' for some reason.

Tested on:
- FreeBSD 10.1-RELEASE

Installed packages:
- postgresql93-client-9.3.6_1
- postgresql93-server-9.3.6 
- python 2.7.9
- used original 'sh' python module (installed with pip)

Steps to reproduce 'hanging':
* check that postgres is stoped
* launch following script:
```python
from sh import sh
from sys import stdout
sh('-x', '-c', '/usr/local/bin/pg_ctl -D /usr/local/pgsql/data start', out=stdout, err=stdout)
```

Conclusion:
The problem source is wrong implementation of daemonization in pg_ctl. Nevertheless we have deadlock in original python sh module when we start such 'false' daemons.
From one hand my fix avoids such situations but in other hand it adds extra call of 'is_alive' method.
Anyway decision about including this 'fix' to sh module is up to you.

amoffat:
Thanks for digging into this Dmytro.  I'll review and this will probably go in the 1.2 release

amoffat:
I believe this issue is related to https://github.com/amoffat/sh/issues/310.  The solution there was to add a timer when the process ends, to shut down the threads processing stdin and stdout.  This should fix your issue as well, since we are guaranteed to break out of the `while` loop at some point after the process dies.  It's on the release-1.2 branch and will ship with it then.  Thanks for reporting your issue and sorry it took so long to finally get to it.

PR #237, opened 2015-03-18
Kill the entire process group when a command  times out

derekhiggins:
In cases where the command being executed spawns subcommands
we need to ensure they are all killed, so that the timeout
occurs when it should instead of waiting for the subprocesses to exit.

The problem I'm trying the fix can be reproduced by setting a timeout on a git clone, git calls subprocesses which prevent the timeout from working. Killing the process group should be safe because os.setsid() was called earlier so the commands have been started in their own group.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -1531,7 +1531,9 @@ class OProc(object):
     def signal(self, sig):
         self.log.debug("sending signal %d", sig)
         try:
-            os.kill(self.pid, sig)
+            # Kill the entire process group as the command may have
+            # spawned subprocesses we don't want left around
+            os.killpg(os.getpgid(self.pid), sig)
         except OSError:
             pass
 


amoffat:
Thanks for reporting this Derek, I'll review and it will probably go with the 1.2 release

apevec:
> go with the 1.2 release

You have 1.11 release, wouldn't 1.2 be lower?

apevec:
Proposed to include as a patch in Fedora/EPEL https://bugzilla.redhat.com/show_bug.cgi?id=1306405

virtuald:
I would propose that instead a `signal_group()` API be added as you might not always want to send any generic signal to the entire group, and would break applications that relied on `signal` sending to just the head process.

I think timeout calling `signal_group()` by default (perhaps with a flag to enable/disable) is fine.

amoffat:
Took @virtuald's suggestion and went the `signal_group()` and `kill_group()` route https://github.com/amoffat/sh/commit/4e8991e100120a7a74b31aba7eb123dc3e986f45.  It's on the `release-1.2` branch and will ship when that goes out soon

PR #241, opened 2015-03-21
Minor update to sh.Command documentation

### LINKED ISSUE #209: Minor: Improve `Command` note in documentation.
*Problem I was trying to solve*
Call `py.test` from within the env. [Documentation](http://amoffat.github.io/sh/index.html#command-execution) states that you need to pass full path to `Command` in case you have dots. This is inconveninet when using many envs, I wondered why `Command` wouldn't resolve a path within an env. It does, only the documentation mislead you to think otherwise. Checked the code and got it working. 

*Solution*
Change in documentation
> ...pass in the absolute path of the executable.

to 

> ...pass in the command name or the absolute path of the executable

mandarvaze:
Specified that absolute path may not be always required. Added an example.
Fixes #209

### DIFF
--- a/_docs_sources/index.rst
+++ b/_docs_sources/index.rst
@@ -79,12 +79,14 @@ substitute the dash for an underscore::
 
     For commands with more exotic characters in their names, like ``.``, or
     if you just don't like the "magic"-ness of dynamic lookups, you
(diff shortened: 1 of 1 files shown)


amoffat:
:+1:

PR #243, opened 2015-03-25
Document the lesser known piped="direct" option

mandarvaze:


snordhausen:
+1 for this pull request. It's a very useful feature, and it would be great to have it properly documented.

marascio:
+1 from me as well. I have multiple chained executions over a large amount of data. Without discovering the `_piped='direct'` they would be impossible. I believe I found it by digging through source.

amoffat:
`_piped=direct` is now the default when `_piped=True` is used, so no documentation needed

PR #245, opened 2015-04-06
Proper behavior with long_sep=' '

zardus:
I was automating some `virt-builder` stuff with `sh`, and `virt-builder` doesn't seem to properly handle long arguments. Specifically, it requires the flag and the value to be separate arguments. As it was, `sh` didn't separate these out properly, and this happened:

```
  STDERR:
/usr/bin/virt-builder: unknown option `--output o/1.img'.
virt-builder: build virtual machine images quickly
```

This pull requests implements argument separation for long arguments with `_long_sep=' '`. The `--flag` and the `value` are simply added as separate arguments.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -857,8 +857,13 @@ output"),
                 elif v is False:
                     pass
                 else:
-                    arg = encode_to_py3bytes_or_py2str("--%s%s%s" % (k, sep, v))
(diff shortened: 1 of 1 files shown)


amoffat:
good stuff.  added a test and it's been merged into the release-1.2 branch and will ship with it

PR #252, opened 2015-05-04
Add test_stdout_file

omribahumi:
Added a test to show a problem with the handling of file objects when passed as a parameter for `_out`

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -1854,6 +1854,9 @@ def get_file_chunk_consumer(handler):
         if hasattr(handler, "flush"):
             handler.flush()
 
+        if hasattr(handler, "close"):
+            handler.close()
+
     return process, finish
 
 def get_callback_chunk_consumer(handler, encoding, decode_errors):
--- a/test.py
+++ b/test.py
@@ -770,6 +770,30 @@ subprocess.Popen(sys.argv[1:], shell=False).wait()
         self.assertEqual(iam1, iam2)
 
 
+    def test_stdout_file(self):
+        py = create_tmp_test(r"""
+import sys
+
+sys.stdout.write("foobar\n")
+""")
+
+        read_fd, write_fd = os.pipe()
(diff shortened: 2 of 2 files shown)


omribahumi:
Ok, seems like the problem only happens with an `os.pipe()`. Apparently, the read end doesn't get an EOF until the file is closed.
It causes problems with other tests, so it's probably not the right fix.

The tests that broke do `seek(0)` on a `TemporaryFile`. We can work around it by passing `os.fdopen(os.dup2(file_obj.fileno()), 'w')` instead of `file_obj` directly, but I guess that's not the expected behavior.
Also, after understanding the true nature of my problem, the test should be named `test_stdout_pipe` and not `test_stdout_file` - in case we decide to keep it.

Any thoughts on this?

amoffat:
I'm against the idea of closing the `file_obj` implicitly.  I think most people would expect it to remain open.  But I am interested into why `.flush()` on the pipe object doesn't finish everything nicely.  Is there an easy way to detect the difference between a pipe object and a file object, and if it's a pipe object, close it?

amoffat:
Ok, what I ended up doing is detecting if the file descriptor is a pipe, and if it is, force close.  Otherwise, leave it open.  https://github.com/amoffat/sh/commit/12bfe9b23d6aedc87720e7c3aebc40b9fb2d44d0 I think this will have the desired effect.  Thanks for the test.  This has been merged into release-1.2 branch and will go out with it then.

amoffat:
@omribahumi I had to roll this back in 1.12.1.  Unfortunately, detecting and closing pipes on process completion caused other problems, like this one https://github.com/amoffat/sh/issues/334

I did add the ability for processes to take raw file descriptors, like pipe fds.  Using raw fds no longer suffer from the flushing problem, so use them in the future:

```python
read_fd, write_fd = os.pipe()
p = python(py.name, _out=write_fd, u=True)
os.read(read_fd, 100) # or a while/read loop
```

PR #259, opened 2015-05-29
support separating long arguments

### LINKED ISSUE #258: Arguments generated from keywords do not work on some programs
For example, `sh.xrandr(output="DP1", off=True)`, which generates `["xrandr", "--output=DP1", "--off"]` does not work because `xrandr` does not understand the argument separated by an equal sign. using `_long_sep=" "` does not work either, because it still is passed as one argument (`["xrandr", "--output DP1", "--off"]`). Some tools, like `xrandr`, require arguments passed in the form `["xrandr", "--output", "DP1", "--off"]`. I suggest either interpreting `_long_sep=" "` as such or adding another configuration variable.

minus7:
…into name and value, fixes #258

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -707,6 +707,8 @@ class Command(object):
         "cwd": None,
 
         # the separator delimiting between a long-argument's name and its value
+        # setting this to None will cause name and value to be two separate
+        # arguments, like for short options
         # for example, --arg=derp, '=' is the long_sep
         "long_sep": "=",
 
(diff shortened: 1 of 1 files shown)


minus7:
Are you going to merge this?
It just adds the option for long args to be passed as separate argvs like with short args when the user explicitly sets _long_sep to None. I think it's a useful addition (and fixes original issue that brought it up, #258).

amoffat:
it's merged into 1.2.  thanks for your work @minus7

Issue #138, opened 2013-05-02
Monitoring the subprocess


NotSqrt:
Hello,

This is more of a feature request, or simply a call for ideas or ways to do this :
I want to follow the resources used by the subprocess I called.

No problem for the wall time:

``` python
import sh, timeit

tic = timeit.default_timer()
sh.sleep(5)
print timeit.default_timer() - tic
```

For memory or CPU time, I'm not interested in a super-precise report, I just thought of something like:

``` python
import time, sh, psutil

cmd = sh.sleep(60, _bg=True)  # have to put it in the background
p = psutil.Process(a.pid)

while p.is_running():  # or cmd.process.alive
    res = (p.get_memory_info(), p.get_cpu_times())
    time.sleep(1)

print res
```

So nothing specific to sh in this.
Do you see any other way to get info about the subprocess ?

Thanks !

PR #262, opened 2015-06-08
Allows baking subclasses of commands into instances of the subclass

NotSqrt:
I subclassed ``sh.Command`` to finally add command monitoring as initially mentioned in amoffat/sh#138, and need this for my subclass to survive baking !


amoffat:
Cool, I see no problems with this.  It'll make it into the next release.  Thanks!

PR #264, opened 2015-06-23
Automatically handle failures in producers of pipe output

hoelzro:
`sh` does not currently raise an exception when the producer side of a pipeline exits with an erroneous exit code unless you explicitly check for it.  This adds tests and fixes.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -479,7 +479,18 @@ class RunningCommand(object):
         # process, and that's if we're using a with-context with our command
         if spawn_process:
             self.log.info("starting process")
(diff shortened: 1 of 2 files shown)


hoelzro:
@amoffat I tried to comply with your coding style as best I could; please let me know if I missed anything.

amoffat:
solid work @hoelzro :+1:   will go out with the 1.2 release

amoffat:
looks like this got fixed during another issue being fixed.  i pulled in the tests though.  thanks for writing those up.

PR #265, opened 2015-07-07
Added PyPy and PyPy3 to the build

thedrow:
There's no reason why they shouldn't be supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,11 @@ language: python
 python:
     - 2.6
     - 2.7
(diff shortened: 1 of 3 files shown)


thedrow:
Strange, it works for PyPy3 but not for PyPy.
Any idea why?

jakirkham:
Appears to be timing failures. May be worth trying the examples in those tests and seeing first that they work and if so how long they take to run. This appears to be an issue with Python 3 too.

Issue #269, opened 2015-08-20
Setting default redirections for all commands


Lucas-C:
Hello,

First, a big THANK YOU for your work on this project.
I find it extremly useful, simple and pythonic.
And it's going to be extremely useful to convince my coworkers to migrate their shell scripts to Python :)

Now, would you be open to allowing some kind of global configuration, specifically on the default redirection for `stderr` ?

Currently, some "hardcoded" defaults are defined there:
https://github.com/amoffat/sh/blob/master/sh.py#L669
And by default the commands `stderr` is simply discarded.

Could that default behaviour be somehow configurable ?
I would like the commands `stderr` to be written to the parent Python script `stderr` by default.
And in the same spirit, it could be sometimes handy to forward `stdout` the same way.

I'd love to work on a pull request if you are ok with this feature request.

Regards

amoffat:
Hi Lucas.  I global configuration sounds useful

In theory, a global configuration is currently possible:

```python
import sh
sh.Command._call_args["timeout"] = 1
sh.sleep(3)
```

This is ugly though because it's manipulating a clearly private variable.  Others have suggested in the past that we could do some kind of configuration object:

```python
from sh import Config
sh2 = Config(timeout=1)
sh2.sleep(3)
```

I think both ways can be acceptable.  In the case of the former way, we should rename `_call_args` to maybe `config` so it's more clear that it is ok to change.

What do you think about implementing a configuration class?  We would also need to add a section to the docs about "Global Configuration" that demonstrates how to use it.  And of course tests :)

Lucas-C:
Sounds like a plan !

What about defining a module global constnat `defaults` and assigning it to `Command._call_args` to keep a full retro-compatibility here ?
We could then do:
```
import sh, sys
sh.defaults['stdout'] = sys.stdout
sh.defaults['stderr'] = sys.stderr
from sh import bzcat, tar, unzip
```
We could implement a configuration class, but if its sole purpose is to provide an access to this global dict of defaults, I'm not sure it's really necessary. I vote for KISS & YAGNI :)
Do you see another use for such class ?

Lucas-C:
After a short code review, it looks to me that `Command._call_args` is only used by:
- [`Command._extract_call_args`](https://github.com/amoffat/sh/blob/master/sh.py#L815)
- [`Command.__call__`](https://github.com/amoffat/sh/blob/master/sh.py#L948)
[...]

PR #271, opened 2015-08-24
Fixing old sh(...) non-documented method to define special args defaults

Lucas-C:
This pull request is an implementation of the changes discussed in #269.

I ended up defining a context manager as its seemed to me the safest and most elegant solution.

Note that this commit removes the old, non-documented, usage to set defaults for all `sh` commands, and its only related test: https://github.com/amoffat/sh/blob/7a4e296/test.py#L1502

The reason beeing that it created a very confusing edge case:
```
import sh, sys
sh = sh(_out=sys.stdout)
from sh import echo  # this command still has _out=None as default
```

I tried many things to work around this issue, but in the end I don't think we can safely allow the `from sh import ...` mecanism AND provide a way to reassign `sh`. It is fundamentally incompatible, due to Python `sys.modules` caching system.

As for the documentation, I couldn't find the source anywhere than on http://amoffat.github.io/sh/, so I generated a patch:
https://gist.github.com/Lucas-C/4f466b14061c5fbfbbd8

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -70,7 +70,7 @@ else:
 
 IS_OSX = platform.system() == "Darwin"
 THIS_DIR = os.path.dirname(os.path.realpath(__file__))
-SH_LOGGER_NAME = "sh"
+SH_LOGGER_NAME = __name__
 
 
 import errno
@@ -153,19 +153,19 @@ class ErrorReturnCode(Exception):
     derived classes with the format: ErrorReturnCode_NNN where NNN is the exit
     code number.  the reason for this is it reduces boiler plate code when
     testing error return codes:
-    
+
         try:
             some_cmd()
         except ErrorReturnCode_12:
             print("couldn't do X")
-            
+
     vs:
         try:
             some_cmd()
         except ErrorReturnCode as e:
             if e.exit_code == 12:
                 print("couldn't do X")
-    
+
     it's not much of a savings, but i believe it makes the code easier to read """
 
     truncate_cap = 750
@@ -350,7 +350,7 @@ class Logger(object):
     script is done.  with sh, it's easy to create loggers with unique names if
     we want our loggers to include our command arguments.  for example, these
     are all unique loggers:
-        
+
             ls -l
             ls -l /tmp
             ls /tmp
@@ -364,7 +364,7 @@ class Logger(object):
         self.name = name
         if context:
             context = context.replace("%", "%%")
-        self.context = context 
+        self.context = context
         self.log = logging.getLogger("%s.%s" % (SH_LOGGER_NAME, name))
 
     def _format_msg(self, msg, *args):
@@ -660,7 +660,7 @@ class Command(object):
     represents the program itself (and not a running instance of it), it should
     hold very little state.  in fact, the only state it does hold is baked
     arguments.
-    
+
     when a Command object is called, the result that is returned is a
     RunningCommand object, which represents the Command put into an execution
     state. """
@@ -787,7 +787,7 @@ output"),
         if not found:
             raise CommandNotFound(path)
 
-        self._path = encode_to_py3bytes_or_py2str(found) 
+        self._path = encode_to_py3bytes_or_py2str(found)
 
         self._partial = False
         self._partial_baked_args = []
@@ -1624,7 +1624,7 @@ class NotYetReadyToRead(Exception): pass
 def determine_how_to_read_input(input_obj):
     """ given some kind of input object, return a function that knows how to
     read chunks of that input object.
-    
+
     each reader function should return a chunk and raise a DoneReadingForever
     exception, or return None, when there's no more data to read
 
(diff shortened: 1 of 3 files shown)


amoffat:
Cool, I'll try to find time this week to review.  The one thing that stands out, and you'll want to verify this for yourself, is I believe this `yield` https://github.com/amoffat/sh/pull/271/files#diff-2c2fab1457aa782cf8f2f34ba9ae0975R2299 needs a `try/finally`, otherwise the environment won't be restored in the case of an exception that bubbles up to the user's code.

Lucas-C:
Good catch, I'll fix that asap.

Also, the `test_done_callback` test seems to be failing in Travis due to a timing issue:
https://travis-ci.org/amoffat/sh/jobs/77056729
Have you seen this kind of test failure before ?

amoffat:
Yeah don't worry about it.  Those Travis machines have very unpredictable timing.

amoffat:
The semantics of the context manager are confusing.  When a context manager is instantiated, typically it is used to use it explicitly to manage state.  The problem I have with this context manager is there is no explicit state:

```python
with sh.defaults(_out=out):
    from sh import echo
```

`sh.defaults` is implicitly changing a global setting on `sh`, but that is not clear to someone reading the code, because context managers typically provide a handle to manage state:

```python
with sh.defaults(_out=out) as sh2:
    from sh2 import echo # new echo
    from sh import echo # old echo
```

If the state is implicit, we still run into the old problem I am trying to avoid of globally changing the behavior of `sh` for another module:

```python
with sh.defaults(_out=out):
    anothermodule.some_function()
```

Where `anothermodule.some_function` has unexpected behavior of the `sh` module.

I want to continue in the direction of explicit state management, so I cannot merge this PR at the moment.

Lucas-C:
Strange, I have this strange behaviour where some `tox` tests fail, but when I run them individually with

    tox FunctionalTests.test_defaults_with_global_module

Or

    python$version test.py FunctionalTests.test_defaults_with_global_module

They pass

Any idea why ??

amoffat:
I haven't tried it yet, but I'm going to guess that there is state bleeding between the tests.  I would look into removing the module from `sys.modules` in the `tearDown` of the test class, and see if that fixes it.  And there doesn't seem to be an existing `tearDown`, so you will have to add one.

Lucas-C:
Despite the Travis checks failing (because of Python 3+ installation issues I think),
this commit looks good to me.
What do you think ?

amoffat:
@Lucas-C lots of tests, that is great to see :)  It will take me awhile to review this..there's quite a bit of code and magic going on, but thanks for the effort you put into the PR.  After I finish reviews, I can see this landing in the 1.2 branch

Lucas-C:
I'm glad to read this !
And sure, take your time :)

sh.py
```diff
@@ -2280,31 +2280,82 @@ def __init__(self, self_module, baked_args={}):
...
 
         # python 3.2 (2.7 and 3.3 work fine) breaks on osx (not ubuntu)
         # if we set this to None.  and 3.3 needs a value for __path__
         self.__path__ = []
         self.__self_module = self_module
-        self.__env = Environment(globals(), baked_args)
-
-    def __setattr__(self, name, value):
-        if hasattr(self, "__env"):
-            self.__env[name] = value
```
amoffat: What was the reason for removing the `__env` stuff?

sh.py
```diff
@@ -2280,31 +2280,82 @@ def __init__(self, self_module, baked_args={}):
...
+        # its class correspond to a previous definition of ModuleImporterFromVariables,
+        # before this file was reloaded
+        if not any(getattr(module, 'restrict_to', None) == self.restrict_to for module in sys.meta_path):
+            sys.meta_path.insert(0, self)
+
+    def find_module(self, mod_fullname, path=None):
+        if self.forbid_mod_names and mod_fullname in self.forbid_mod_names:
+            return None
+        parent_frame = inspect.currentframe().f_back
+        while parent_frame.f_code.co_filename == '<frozen importlib._bootstrap>':
```
amoffat: Could you add a comment about what this line is for?  My rough understanding is that the frozen modules are inside the importlib itself?  And we're trying to get the parent frame until we're "out" of the importlib code?

test.py
```diff
@@ -4,8 +4,8 @@
 from os.path import exists, join, realpath
 import unittest
 import tempfile
+import sh as _sh # importing under another name to avoid skewing tests
```
amoffat: smart :+1:

sh.py
Lucas-C: I did not remove the `__env`, just the `__setattr__` magic method.
The reason is that I did not find any use case for it, and all the tests passed ok without it.
So I just removed it to simplify the code.

sh.py
Lucas-C: Which line exactly ? The `while` in `find_module` ?
Then yes, you got the reason right. I'll add a comment

amoffat:
Apologies for being so slow on this.

I actually have some really bad news about this PR.  The last few days, I was tinkering with it (just to understand it better), and I noticed the tests fail on python3.3 and python3.4 (but work in 3.2, 3.1 and 3).  It turns out that in python3.3 and later, the internal import code *requires* any module that you load with a fancy importer (like we've done here) be put into `sys.modules`.

However, as your docstring on `load_module` states, we deliberately don't put the module in `sys.modules`, the reason being that if it's cached, this no longer works:

```python
import sh

sh1 = sh(_out="/dev/null")
from sh1 import ls
print(ls()) # shouldn't print

sh1 = sh()
from sh1 import ls
print(ls()) # SHOULD print, but doesn't!
```

`sh1` has been cached in `sys.modules`, and so the second `from sh1 import ls` re-uses that existing module, with the old baked (_out="/dev/null") params.

There really doesn't seem to be a way around this.  Python 3.3 and up require the module in `sys.modules`, and if it's in there, tests fail and behavior breaks.  I think best-case scenario is we remove the `from / import` and just go with using attribute lookups, like `sh1.ls()` instead.  This is a real bummer because of the amount of work that went into the discussion and implementation.  I'm open to any ideas.

amoffat:
Actually, scratch that last message.  I think I found a way around the limitation, but it's horrible and I feel bad for writing it.  But it works!



It allows us to put our module in `sys.modules`, and then pre-preemptively remove it if another context would try to re-use it

Lucas-C:
Glad you found the time to get back to this PR ! ^^
OMG, yes, this is very hacky. I wish we wouldn't need to do such horrible stack frame manipulation...
So, this will be released in 1.2 ?

amoffat:
It will, yes.  I am aiming for some time next week if I can make it through all the issues and PRs by then

PR #276, opened 2015-09-14
cd allows no arguments to move to the home directory

rsmith31415:
```
>>> sh.cd('test')
>>> sh.pwd()
/home/user/test
>>> sh.cd()
>>> sh.pwd()
/home/user
```


amoffat:
looks good.  not sure why all the mac tests are failing, but i think that's travis's fault

rsmith31415:
I think that is likely. It fails when creating the environment.

amoffat:
this is on the release-1.2 branch and will go out with it.  thanks for your contribution

Issue #283, opened 2015-11-05
ErrorReturnCode not well documented


jorisroovers:
I was looking at the documentation on [http://amoffat.github.io/sh/#exit-codes]() and found that the  `ErrorReturnCode` exception is not well documented.
In particular, I found that the following members of the exception are not mentioned: `args, exit_code, full_cmd, message, stderr, stdout, truncate_cap`.

I'd be happy to submit a pull request that adds a little more detail if this issue is confirmed.

amoffat:
Confirmed, some more docs would be welcome @jorisroovers

amoffat:
http://amoffat.github.io/sh/sections/command_class.html#errorreturncode

PR #284, opened 2015-11-05
Added details about ErrorReturnCode

jorisroovers:
Added more details about the ErrorReturnCode exception to the Exit Codes
section. This is for #283.


imZack:
:+1: this makes docs more clear

PR #290, opened 2015-12-11
Actually change directory in pushd context

ppannuto:
The pushd context needs to actually change the directory instead of just
setting the _cwd argument. If it doesn't it leads to unexpected behavoir
such as `sh.glob` and `os.getcwd` not working correctly inside of the
pushd context.

This also changes pushd to restore the directory the user was in when
exiting the context, even if the user cd's while inside of the context.
This mirror's bash's pushd/popd behavoir.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -28,7 +28,6 @@ __version__ = "1.11"
 __project_url__ = "https://github.com/amoffat/sh"
 
 
-
 import platform
 
 if "windows" in platform.system().lower():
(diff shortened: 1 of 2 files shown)


ppannuto:
I looked into the CI failures, and they all look like build system errors, independent of this patch. For configurations that the build system works, all tests run successfully.

amoffat:
@ppannuto good work on this.  i made some alterations, including adding thread safety with a re-entrant lock, and it's been merged into the release-1.2 branch and will go out with it then.

PR #292, opened 2016-01-07
handle EINTR by retrying waitpid (bug in python < 3.5)

mjmvisser:
waitpid can fail with errno set to EINTR if an unblocked signal or SIGCHLD is caught (see http://www.tutorialspoint.com/unix_system_calls/waitpid.htm). Prior to python 3.5, this must be explicitly detected and the waitpid call retried (see https://www.python.org/dev/peps/pep-0475/).

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -49,6 +49,7 @@ from types import ModuleType
 from functools import partial
 import inspect
 from contextlib import contextmanager
+import errno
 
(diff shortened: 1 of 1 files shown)


amoffat:
Thanks for the contribution @mjmvisser.  Your change + a test are on the release-1.2 branch and will go out when it ships.

PR #311, opened 2016-04-14
Allow configuration of the log message

novas0x2a:


test.py
```diff
@@ -1923,6 +1923,31 @@ def test_signal_exception_aliases(self):
...
+            logger.addHandler(handler)
+            with sh.args(log_msg=log_msg):
+                python(py.name, "meow", "bark")
+        finally:
+            logger.removeHandler(handler)
+
+        loglines = buf.getvalue().split('\n')
+        self.assertTrue(loglines, 'Log handler captured no messages?')
+        self.assertTrue(
+            fnmatch.fnmatch(loglines[0],
```
amoffat: fnmatch confused me, but I guess it is because `<Command` will contain the temp file path?

amoffat:
Hi @novas0x2a, could you elaborate a bit on the use case for this?  Is it to make all logging lines prefixed with a custom value?  And I am guessing it cannot be done by modifying a logger retrieved from `logging.getLogger` because we use a custom `Logger` class wrapper?

novas0x2a:
Hi! Yeah, no problem: I'm using it mostly to turn off the truncation, in general I want to see the full command as exec'd in the logs (including all args), and then only call_args that are non-default; i figured doing it this way might be valuable if others wanted something different.

amoffat:
Aha, I get it.  Ok looks good, I will make sure it gets into the next release.  :+1: Don't worry about the tests failing.. travis acting up.

test.py
novas0x2a: Yeah, that was why, and it neatly works around the `python` vs `python3.5` problem, too.

novas0x2a:
Great! :D

amoffat:
it's on the release-1.2 branch.  thanks for the contribution @novas0x2a.  also to note, the base logging was changed to be more like what you described you wanted (non-truncated command, fewer call args), so you may not even need `log_msg` when 1.2 goes out.  good to have in any case

PR #312, opened 2016-05-03
Support univeral wheels

borgstrom:
sh already supports both py2/py3 from the same code base, so there's no
reason not to produce a universal wheel.

We need to try and import setuptools first (so we can even build wheels
in the first place) and fallback to distutils only if we don't have
setuptools.  (We should prefer setuptools anyways:
https://python-packaging-user-guide.readthedocs.io/en/latest/current/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,9 @@ env:
     - LC_ALL=C
     - LC_ALL=en_US.UTF-8
 
-before_script: pip install coverage coveralls
+before_script:
+    - if [[ $TRAVIS_PYTHON_VERSION == 3.2 ]]; then pip install 'coverage<4'; fi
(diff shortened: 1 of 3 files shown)


borgstrom:
Not sure what's up with the OS X travis jobs, but this also fixes the py3.2 issue with coverage now.  See travis-ci/travis-ci#4866

borgstrom:
@amoffat ping.

amoffat:
Thanks for your interest.  Development on sh is not dead, but I currently do not have time for active development.  As my schedule clears up, I will resume development.  Apologies if you have opened a PR or contributed code and are waiting for that code to be reviewed, tested, and merged.

amoffat:
wheel is up!

PR #316, opened 2016-06-16
Removed encoding/decoding mess in line-buffered mode

ecederstrand:
Despite the comment, keeping chunk and chunk_to_write in bytes works fine. Tested in Python2.7 and Python3.4 with no (additional) tests failing. Unfortunately, I can't read from the comments what the original concern was.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -2052,34 +2052,26 @@ class StreamBufferer(object):
                 return [chunk]
 
             # line buffered
-            # we must decode the bytes before we try to match on newline
             elif self.type == 1:
                 total_to_write = []
-                chunk = chunk.decode(self.encoding, self.decode_errors)
+                nl = "\n".encode(self.encoding)
                 while True:
-                    newline = chunk.find("\n")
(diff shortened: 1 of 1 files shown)


ecederstrand:
A lot of the tests seem to be failing for completely unrelated reasons ("tar: Unrecognized archive format" and SyntaxError in the "coverage" package)

amoffat:
This does appear to be simpler and functionally identical.  I'm seeing some of the Travis 2.7 tests fail though:

ecederstrand:
Yes, that's really bizarre. Python2.7 is complaining that the test file for `FunctionalTests.test_decode_error_handling` doesn't contain the "`# -*- coding: utf8 -*-`" stanza, when that is clearly the case, according to `test.py`.

The test doesn't fail on my own machine (FreeBSD 10.1):

```
$ python2.7 test.py FunctionalTests.test_decode_error_handling
.
Ran 1 test in 0.369s

OK
```

amoffat:
merged into `release-1.2`  great work

PR #326, opened 2016-09-07
add append option

rains31:
we can't keep the history content because it clears the output file every time. 
but the history contents may be useful as well when using for logging something.


amoffat:
@rains31 Passing the filename to `_out` is more for just convenience, but if you want more flexibility, pass a file-like object to `_out`.  I have demonstrated this with a test: https://github.com/amoffat/sh/commit/0b85d24006b02c828d7091ca7fd16746c711aa78

PR #327, opened 2016-09-23
Spelling Correction: s/interable/iterable/, and then some

james-prior:
Fixed spelling from interable to iterable.

Also created requirements-docs.txt and updated README.md to use it.


james-prior:
On Fri, 23 Sep 2016 18:57:40 -0700, Andrew Moffat <notifications@github.com> wrote:

> Hey @james-prior, could you explain the html and js changes a bit?

That's what I got from running `python compile_docs.py 1.11`
per README.md.

> Are they from an upgraded sphinx?

I suspected so, although I did not _know_ so.
I documented which version of sphinx I used in requirements-docs.txt.
I did not know which version of sphinx was used before me.
I was startled by how many changes there were after 
running compile_docs.py.
Casual browsing of the results looked OK to me.

Reviewing the differences between commits 89cad12 and b2b7fbc,
reveals that the docs in 89cad12 were created with sphinx 1.3.1.
I went back to 89cad12, created a new branch, gh-pages-sphinx-1.3.1,
and recompiled the docs with sphinx 1.3.1. That yielded many small
differences, but no where as many as with sphinx 1.4.6,
which yields many big changes, as you likely saw.
You can see my experimental gh-pages-sphinx-1.3.1 branch at
https://github.com/james-prior/sh/tree/gh-pages-sphinx-1.3.1
I wonder if the differences from c58e257 to c0c1c01
are due to version differences for the other packages used.

If you want particular versions of sphinx or other packages
to be used, please specify them in requirements-docs.txt.

amoffat:
@james-prior no worries, thanks for the clarity and thank you for the PR

amoffat:
thanks.  it's been merged into the next docs release branch

PR #337, opened 2016-11-29
Fix memory leak when running commands

tronical:
In commit dd571459719473a00c143eb37b7661207c5ff9ef the file descriptor
closure of session_pipe_read was accidentally removed, causing a leak.

The test at least on Linux didn't detect this because the lsof output
prints "pipe" in lower-case, instead of upper case, as expected by the
test. Therefore the fd leak test didn't account for the leaked
descriptors.


amoffat:
merged :+1:

PR #342, opened 2016-12-01
don't actually raise a DeprecationWarning

scottbelden:
If you raise the DeprecationWarning then it acts as an actual exception.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -104,6 +104,7 @@ import resource
 from collections import deque
 import logging
 import weakref
+import warnings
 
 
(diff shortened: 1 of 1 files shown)


amoffat:
It's intentional.  `DeprecationWarning` is a subclass of `Exception`, and people shouldn't be using `sh.args` anymore due to its lack of safety.

scottbelden:
I see, it was actually removed. It still seems that this should be a different exception. I'm probably not the only person who was surprised that a warning was really an error.

PR #348, opened 2016-12-15
Make glob patch support `recursive` argument in Python 3.5

### LINKED ISSUE #341: glob monkey patch missing `recursive` argument in Python 3.5
Hi, in your newest release the "monkey patch" of `glob.glob` breaks user code, if one tries to set the `recursive` argument that the original `glob.glob` implementation provides in Python 3.5 ([see Python 3.5 docs](https://docs.python.org/3/library/glob.html#glob.glob)).

Demonstration:
```python
#> python3.5
>>> import glob
>>> glob.glob('.', recursive=True)
['.']
>>> import sh
>>> glob.glob('.', recursive=True)
TypeError: glob() got an unexpected keyword argument 'recursive'
```

I guess adding `args` and `kwargs` to your `glob` wrapper might fix this and makes you independent of the function signature of the actual `glob` implementation. Suggested diff on `sh.py`:
```diff
377,378c377,378
< def glob(path):
<     expanded = GlobResults(path, _old_glob(path))
---
> def glob(path, *args, **kwargs):
>     expanded = GlobResults(path, _old_glob(path, *args, **kwargs))
```

Many thanks for your great work! ;)

*edit: fixed order in the diff*

elritsch:
Modification to the `glob` patch to make it forward any `*args` and `**kwargs` to the original `glob` implementation. This adds support for the newly added `recursive` argument in Python 3.5.

See Python 3.5 `glob` documentation:
https://docs.python.org/3/library/glob.html#glob.glob

Resolves #341

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -375,8 +375,8 @@ class GlobResults(list):
         self.path = path
         list.__init__(self, results)
 
-def glob(path):
-    expanded = GlobResults(path, _old_glob(path))
+def glob(path, *args, **kwargs):
+    expanded = GlobResults(path, _old_glob(path, *args, **kwargs))
     return expanded
 
 glob_module.glob = glob
--- a/test.py
+++ b/test.py
@@ -266,6 +266,17 @@ print(sys.argv[1:])
         out = python(py.name, files).strip()
         self.assertEqual(out, "['*.faowjefoajweofj']")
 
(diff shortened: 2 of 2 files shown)


amoffat:
Good work @elritsch, thanks for taking care of this :+1:  Your fix is live on version 1.12.8

PR #364, opened 2017-02-27
try to fix issues 356

zhangyafeikimi:
Please test it carefully.
I only test it under ubuntu 16.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -130,6 +130,132 @@ if IS_PY3:
 _unicode_methods = set(dir(unicode()))
 
 
+POLLER_EVENT_READ = 1
+POLLER_EVENT_WRITE = 2
+POLLER_EVENT_HUP = 4
+POLLER_EVENT_ERROR = 8
+if hasattr(select, "poll"):
+    class Poller(object):
+        def __init__(self):
+            self._poll = select.poll()
+            # file descriptor <-> file object bidirectional maps
+            self.fd_lookup = {}
+            self.fo_lookup = {}
+
+        def __nonzero__(self):
+            return len(self.fd_lookup) != 0
+
+        def __len__(self):
+            return len(self.fd_lookup)
+
+        def _set_fileobject(self, f):
+            if hasattr(f, "fileno"):
+                fd = f.fileno()
+                self.fd_lookup[fd] = f
+                self.fo_lookup[f] = fd
+            else:
+                self.fd_lookup[f] = f
+                self.fo_lookup[f] = f
+
+        def _remove_fileobject(self, f):
+            if hasattr(f, "fileno"):
+                fd = f.fileno()
+                del self.fd_lookup[fd]
+                del self.fo_lookup[f]
+            else:
+                del self.fd_lookup[f]
+                del self.fo_lookup[f]
+
+        def _get_file_descriptor(self, f):
+            return self.fo_lookup.get(f)
+
+        def _get_file_object(self, fd):
+            return self.fd_lookup.get(fd)
+
+        def _register(self, f, events):
+            # f can be a file descriptor or file object
+            self._set_fileobject(f)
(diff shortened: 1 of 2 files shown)


amoffat:
@zhangyafeikimi can you make your changes to this PR please?  https://github.com/amoffat/sh/pull/362/files#diff-2c2fab1457aa782cf8f2f34ba9ae0975R132

We need `poll_select` to behave exactly like `select.select`, so we can use one or the other, depending on if the system supports `select.poll` or not.

zhangyafeikimi:
Actually, poll_select to hardly behaves exactly like select.select. I don't  think it is the right way.
For example, the 1st and 2nd our_select are running in a while loop, an intrusive modification is better, using poll_select instead of select.select is less efficient.


As I said early, all posix OS have poll, and select has a fatal deficiency.
poll is always preferred, no need to switch between select or poll.

zhangyafeikimi:
@amoffat 
If you insist on switch between select and poll, an intrusive modification is also better.
Don't try to expect poll_select to be equivalent to select.select.

amoffat:
I don't think the calling code should care whether select.select or select.poll was used, that's why it's abstracted out into a separate function, poll_select.  In order for that to work though, it has to behave the same.  I don't see any reason why it shouldn't be able to behave the same.  The only problem I am interested in addressing is the file-descriptor limit with select.select, so I don't care if we are not using select.poll efficiently.

I do insist on using poll_select because it is not fully clear which systems in the wild [HAVE_BROKEN_POLL](https://github.com/python/cpython/blob/6f0eb93183519024cb360162bdd81b9faec97ba6/Modules/selectmodule.c#L329) and I'm not interested in breaking those systems and then coming back to this problem.

zhangyafeikimi:
Yes, "HAVE_BROKEN_POLL" makes sense.
Well, don't try to let poll be like select. They have something different in nature.
Let's make another abstract layer to adapt select or poll or epoll or others...

amoffat:
> Let's make another abstract layer to adapt select or poll
I agree, that is what I was trying to do with `poll_select` in [this PR](https://github.com/amoffat/sh/pull/362/files).  Unfortunately I cannot merge your PR until I am confident that it will work on systems where `select.poll` is broken or unimplemented.

zhangyafeikimi:
@amoffat Please don't merge my PR, it only uses poll.
I will try to make a abstract layer in days.

amoffat:
Ok sounds good @zhangyafeikimi.  Reach out to me if you run into troubles.

zhangyafeikimi:
updated

amoffat:
good work @zhangyafeikimi.  very clean abstractions :+1: .  I merged your code into [release-1.12.10](https://github.com/amoffat/sh/pull/367/commits/8a21e8785866db612979fb84a02264ff86daf1ae) and added a formal test.  it will be merged into master shortly

zhangyafeikimi:
See it. Thank you.

boksic1986:
how does sh use wildcase? like 
`proc = subprocess.Popen('ls *.bc', shell=True)`

amoffat:
@boksic1986 This isn't the correct issue for your question, but this should help http://amoffat.github.io/sh/sections/faq.html#why-doesn-t-work-as-a-command-argument

boksic1986:
Thanks~：）



--
Chen Jiucheng

College of Animal Science , Zhejiang University

No.866 Yuhangtang RD. ,Hangzhou,  China

Tel:0571-88982308
 
From: Andrew Moffat
Date: 2017-06-19 14:53
To: amoffat/sh
CC: boksic1986; Mention
Subject: Re: [amoffat/sh] try to fix issues 356 (#364)
@boksic1986 This isn't the correct issue for your question, but this should help http://amoffat.github.io/sh/sections/faq.html#why-doesn-t-work-as-a-command-argument

PR #369, opened 2017-03-09
Explicitly add support for Python 3.6

samueldg:
- Updated the automated tests to run on Python 3.6;
- Added the version the PyPI classifiers;
- Edited the docs to indicate support up to 3.6;
- Edited a comment about inconsistent ctrl-d behavior in Python
  (tested 3.6 behaves like 3.3-3.5)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ python:
     - 3.3
     - 3.4
     - 3.5
+    - 3.6
     - pypy3
 
 before_script:
--- a/README.rst
+++ b/README.rst
@@ -13,7 +13,7 @@
 
 |
 
-sh is a full-fledged subprocess replacement for Python 2.6 - 3.5, PyPy and PyPy3
+sh is a full-fledged subprocess replacement for Python 2.6 - 3.6, PyPy and PyPy3
 that allows you to call any program as if it were a function:
 
 .. code:: python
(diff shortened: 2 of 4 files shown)


samueldg:
@amoffat As you can see, one of the test is failing:
https://travis-ci.org/samueldg/sh/jobs/209206256

`FunctionalTests.test_piped_exception1` fails for "Python3.6, locale 'C', poller: poll".

I'm testing on Mac OSX though, and the tests, including this one, are all passing for Python 3.6.
I do however have weird message that Python unexpectedly shut down, which could be related (?)

amoffat:
Thanks for reporting this @samueldg, I will try to investigate tomorrow

amoffat:
I refreshed the build on Travis and everything looks good.  Running the tests locally on linux looks good too.  The 120 error code from the [errored build](https://travis-ci.org/samueldg/sh/jobs/209206256#L838) is interesting, and leads [here](https://bugs.python.org/issue29130).  I have seen some weirdness in the tests as well from time to time.

I will merge this regardless, since I think the problem is isolated to the tests, and cannot be reproduced consistently.

samueldg:
Interesting! Thanks for the quick follow-up, and for the amazing lib.

amoffat:
Thank you for helping it get to python 3.6 :+1:

PR #377, opened 2017-03-30
Fix RST error in README.rst

Flimm:
This will make the PyPI page for sh look pretty again.

This includes a test to make it harder to accidentally cause this kind of error again.


amoffat:
@Flimm so much nicer, good work.  Anonymous rst hyperlinks..learned something new today.  Also thank you for the test :+1:

Issue #398, opened 2017-07-18
Importing sh excepts on further imports via Python C-API


cheriff:
I've found that when writing a C Application against libpython that:

1. Use PyImport_ImportModule to import (directly or indirectly) sh
    * This will insert a ModuleImporterFromVariables in sys.meta_path
2. Try use PyImport_ImportModule to import anything else (say, datetime)

This results in:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/sh.py", line 3421, in find_module
    while in_importlib(parent_frame):
  File "/usr/lib/python2.7/site-packages/sh.py", line 3368, in in_importlib
    return frame.f_code.co_filename == "<frozen importlib._bootstrap>"
AttributeError: 'NoneType' object has no attribute 'f_code'
```

After digging, it seems that find_module() is not happy when parent_frame is None, as it appears to be when  importing via the API. I found that having that return None in this case causes things to work, but cannot be sure if this doesn't have side-effects should anyone be using the "fancy importer"

---

In this situation, since I can't easily change either sh itself, or dependancies that pull it in I do have a workaround which I'm using for now:

```
    {   // begin workaround
        PyObject *sys = PyImport_ImportModule("sys");
        PyObject *meta_path = PyObject_GetAttrString(sys, "meta_path");
        PyObject *sh = PyImport_ImportModule("sh");
        PyObject* pop_ret = PyObject_CallMethod(meta_path, "pop", NULL);
        (void)pop_ret;
        (void)sh;
    }
```

I now can import my other stuff since there's no extra importers in the chain, and since sh is already imported, one won't be re-added. 
I understand this may break uses of sh in any module, particularly if anything uses the `my_sh = sh(_foo=bar); from my_sh import something` pattern which for now seems not to be a deal breaker but feels a bit fragile going forward.

---

Alternatively, if there's anything extra I need to do from the C side to set things up such that `inspect.currentframe().f_back` returns something sensible to also avoid this, I'd be happy to try that too.

cheriff:
Repro code, with optional workaround:

```
12:10:37 [davidm@krill ~]$ cat importsh.c
#include <Python.h>
#include <stdio.h>

void workaround(void)
{
    PyObject *sys = PyImport_ImportModule("sys");
    PyObject *meta_path = PyObject_GetAttrString(sys, "meta_path");
    PyObject *sh = PyImport_ImportModule("sh");
[...]

PR #400, opened 2017-07-21
(potential) Fix for issue #398.

davidm-mm:
Give up on trying to fancy import if there is no parent in the call stack.
This seems to be the case when calling PyImport_ImportModule() from C code,
and if sh has already added itself to sys.meta_path exceptions would be thrown
on subsequent imports.

This fixes my simple test case: i.e calling workaround() is no longer required for the C snippet in #398 to work.

The importlib stuff is a bit magic for me, but I dont *think* this would break existing uses because:
* If parent was ever None before now, things would not be working anyway
* the test suite still passes (on my centos box)

I'd be happy for this to be addressed in another way if preferred, but this would scratch my itch as-is.

Thanks,
Dave M

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -3430,6 +3430,13 @@ class ModuleImporterFromVariables(object):
         the rest of this function """
 
         parent_frame = inspect.currentframe().f_back
+
+        # Calling PyImport_ImportModule("some_module"); via the C API may not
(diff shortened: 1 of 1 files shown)


amoffat:
Closing in favor of https://github.com/amoffat/sh/pull/494, which builds on this

PR #416, opened 2018-01-24
Add optional timeout parameter to RunningCommand

lgrahl:
I'm not sure if `timeout` would need to be passed to `self.process._stdin_process.command.wait()` too.


lgrahl:
Right... this obviously will not work. :)

PR #430, opened 2018-04-16
Long live ash!

nierob:
Ash allows to use sh through asyncio event loop.

By default all sh calls are blocking, unless one uses a _bg option
and _done callbacks which produces code a bit hard to maintain.

Ash implementation sets _bg, _bg_exc, _new_session and _done, there
should be no point in providing these for a user.
 - _bg is used to avoid blocking
 - default _bg_exc is wrong for this use case as we always wait
   unless the operation is cancelled
 - default _new_session good for interactive coding, but when running
   in an event loop we really do not want to leak subprocesses
 - there should not be any need for _done callback as await syntax
   is superior and callbacks can be handled by asyncio.Future

Example of usage:
```python
import ash
# collects paths to delete
paths = await ash.ls()
# delete them in parallel
delete_coroutines = (ash.rm(path) for path in paths)
await asyncio.gather(*delete_coroutines)
```

### DIFF
--- a/ash.py
+++ b/ash.py
@@ -1,0 +1,63 @@
+"""
+http://amoffat.github.io/sh/
+"""
+#===============================================================================
+# Copyright (C) 2018 The Qt Company Ltd.
+# Contact: http://www.qt-project.org/legal
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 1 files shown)


nierob:
I guess the patch is not well integrated with the rest of code, but I would need some hints how to do it. That class "works for me"

ash.py
```diff
@@ -0,0 +1,56 @@
...
+    async def __call__(self, *args, **kwargs):
+        loop = asyncio.get_event_loop()
+        lock = asyncio.Lock(loop=loop)
+        await lock.acquire()
+        proc = self.command(*args, **kwargs, _bg=True, _bg_exc=False, _new_session=False, _done=lambda *args: loop.call_soon_threadsafe(lock.release))
+        try:
+            await lock.acquire()
+        except:
+            # The operation was probably cancelled, the subprocess is not needed anymore
+            proc.terminate()
```
nierob: oops there is a race condition between an exception and termination of the process, I will update the code

nierob:
Ping? Any hints on how to improve / merge it?

nierob:
In the end I have written code with sh compatible api that uses asyncio directly. It is technically superior then wrapping sh.

tony:
> In the end I have written code with sh compatible api that uses asyncio directly. It is technically superior then wrapping sh.

I know it's been a while, but this is interests me

If you have any code examples or further work it'd be interesting to see

if anyone else has further issues to link to, or asyncio `sh` examples, this is interesting to hear about for those that find this thread

tony:
> I feel bad for missing this thread back when it was first opened. FWIW, there is asyncio support in sh v2.0.0. It just needs a couple more tests around it, but I believe it is complete.

I suppose this is happening in #585?

Looking forward to it.

Issue #347, opened 2016-12-12
tests assume existence of some binaries


0-wiz-0:
The tests assume that `python` and `lsof` exist.
Some distributions (like pkgsrc) only provide versioned python binaries, i.e. `python3.5`; and `lsof` may or may not be installed.

amoffat:
I understand `lsof` not existing, but could you elaborate on python not existing as well?  The tests should only expect the python version that the test is currently running under.

0-wiz-0:
Most tests are fine in that regard, but some not. Here's the complete error log from the end of the test run on NetBSD 7.99.42/amd64 with python-3.5.2:
````
======================================================================
ERROR: test_importer_detects_module_name (test.ExecutionContextTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File ".../sh-1.12.7/test.py", line 2865, in test_importer_detects_module_name
    from omg import python
ImportError: cannot import name 'python'

======================================================================
ERROR: test_background_exception (test.FunctionalTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File ".../sh-1.12.7/test.py", line 821, in test_background_exception
    self.assertRaises(exc_to_test, p.wait) # should raise
  File "/usr/pkg/lib/python3.5/unittest/case.py", line 727, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/usr/pkg/lib/python3.5/unittest/case.py", line 176, in handle
    callable_obj(*args, **kwargs)
  File ".../sh-1.12.7/sh.py", line 656, in wait
    self.handle_command_exit_code(exit_code)
  File ".../sh-1.12.7/sh.py", line 677, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: 

  RAN: .../.tools/bin/ls /ofawjeofj

  STDOUT:


  STDERR:
ls: /ofawjeofj: No such file or directory


======================================================================
ERROR: test_fg (test.FunctionalTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File ".../sh-1.12.7/test.py", line 1550, in test_fg
    sh.python(py.name, _fg=True)
  File ".../sh-1.12.7/sh.py", line 3172, in __getattr__
    return self.__env[name]
  File ".../sh-1.12.7/sh.py", line 3049, in __getitem__
    raise CommandNotFound(k)
sh.CommandNotFound: python

======================================================================
ERROR: test_general_signal (test.FunctionalTests)
[...]

PR #438, opened 2018-06-20
always use fully versioned python command in tests

carlwgeorge:
I maintain python-sh in Fedora.  [Fedora is working towards deprecating the unversioned `python` command](https://fedoraproject.org/wiki/Changes/Avoid_usr_bin_python_in_RPM_Build).  This change will update the test suite to always use `python{major}.{minor}`, and not assume the existence of `python` or even `python{major}`.

### DIFF
--- a/test.py
+++ b/test.py
@@ -93,22 +93,21 @@ if IS_PY3:
     ioStringIO = StringIO
     from io import BytesIO as cStringIO
     iocStringIO = cStringIO
-    python = sh.Command(sh.which("python%d.%d" % sys.version_info[:2]))
 else:
     from StringIO import StringIO
     from cStringIO import StringIO as cStringIO
     from io import StringIO as ioStringIO
     from io import BytesIO as iocStringIO
-    python = sh.python
 
 THIS_DIR = dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


hroncok:
Maybe use sys.executable instead?

carlwgeorge:
That's a good point, that would be cleaner.  For this PR I kept the previous method, but could adjust.

@amoffat, which one would you prefer?

* `system_python = sh.Command(sh.which("python%d.%d" % sys.version_info[:2]))`
* `system_python = sh.Command(sys.executable)`

hroncok:
In most casesy, this behaves the same, but not in all. sys.executable is safer here. You don't necessarily have the versioned executable on PATH.

carlwgeorge:
I updated this PR with @hroncok's suggestion.  How does this look to you @amoffat?

carlwgeorge:
I added a few more fixes to this pull request to address the effects of `/usr/bin/python` not existing (see also #347).  @amoffat, could you take a look and consider merging this?

ecederstrand:
@amoffat I think this is a useful improvement. I don't see how we can reasonably add test coverage.

amoffat:
LGTM :+1: It will go out with the next release.

Issue #446, opened 2018-08-13
`TimeoutException` does not subclasses 


honnix:
`TimeoutException` doesn't seem to subclass `ErrorReturnCode` as documented, see https://github.com/amoffat/sh/blob/master/sh.py#L394

Would be good to make it a subclass of `ErrorReturnCode` so that when catching this exception, `full_cmd` is available for logging purpose.

ecederstrand:
@amoffat Would you prefer to fix the code to inherit from `ErrorReturnCode` (see https://github.com/amoffat/sh/blob/develop/sh.py#L406), or fix the documentation (see https://github.com/amoffat/sh/blob/gh-pages/_docs_sources/sections/command_class.rst#timeoutexception)?

amoffat:
With https://github.com/amoffat/sh/pull/447 merged, I say it's probably "good enough" to just fix the docs.  Do you want to take that one @ecederstrand?  There is a [gh-pages-release-1.12.15](https://github.com/amoffat/sh/tree/gh-pages-release-1.12.15) branch I've been adding to for the next release.

ecederstrand:
I agree. A process timeout is not a return code, since the process did not exit by itself.

PR #447, opened 2018-08-13
capture full command in TimeoutException

honnix:
To address #446

PTAL @amoffat


ulzha:
:+1:

amoffat:
Seems reasonable :+1: Will go out with the next release

Issue #422, opened 2018-03-02
*** TypeError: 'NoneType' object is not subscriptable


davisem:
Trying to use the _env parameter (very cool btw), but am seeing this error:
```
Python 3.6.2 (default, Jul 17 2017, 16:44:45)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sh
>>> from io import StringIO
>>>
>>> buf = StringIO()
>>> sh2 = sh(_out=buf)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/sh.py", line 3367, in __call__
    code = parent[4][0].strip()
TypeError: 'NoneType' object is not subscriptable
```

I was also hoping I could do something like:
```
>>> my_env = {'PATH': 'somepath'}
>>> sh2 = sh(_env=my_env)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/sh.py", line 3367, in __call__
    code = parent[4][0].strip()
TypeError: 'NoneType' object is not subscriptable
```

amoffat:
Hmm..it looks like you've caught an edge case from some XTREME sh magic being used.  I bet what you're doing will work in a python script file, even if it doesn't work on the python console.  Give it a try.

davisem:
Truly magical. Thank you for you help, and for the awesome module. Long live sh!

amoffat:
Np.  Let's leave this issue open until I fix the underlying bug :+1:

ddstreet:
@amoffat are you still maintaining this pkg?  This still isn't fixed...

PR #466, opened 2019-03-02
Fix execution contexts on the REPL

abadger:
This is a fix for #422 

On the REPL and from the command line, Python doesn't keep source code
in the stack frame for the toplevel.  Therefore we can't inspect it and
look for the variable being assigned to in order to remove it from
sys.modules.

This is a cornercase but we can hack around it by trying to retrieve the
variable name from the code object instead.  This depends more heavily
on Python internals but this was already code which dove into Python's
internals a bit so perhaps this is okay.

<strike>Since I needed to run the tests on Python3.7, this also includes a fix for the test suite on Python3.7</strike>  I pulled the Python-3.7 changes out to their own PR.


abadger:
I'm not sure what the issue is trying to build this on pypy.  The test suite passes locally for me on pypy but I have a newer version than is running on travis:
```
$ pypy --version
Python 2.7.13 (1d8462e42c37, Aug 30 2018, 09:58:57)
[PyPy 6.0.0 with GCC 8.2.1 20180801 (Red Hat 8.2.1-2)]
```
https://toshio.fedorapeople.org/sh-passing-tests-on-local-pypy.log

abadger:
Okay, I grabbed the travis version of pypy in a VM and I found the problem.  Older versions of pypy don't report parent[1] in the same way as CPython (and newer pypy corrects it).  I'll special case that to get it to work.

amoffat:
Great work on this @abadger and solid test :+1:.  Will go out with the next release.

abadger:
Cool.  Thanks for merging all the outstanding PRs :-)

I noticed some sloppy coding I included in this PR.  I'll submit another PR within the hour to address that.

PR #467, opened 2019-03-02
Raising exceptions should use native strings.

### LINKED ISSUE #463: Python 2: Non-ASCII characters in stderr or stdout during non-zero exit result in exception with no message
The `sh` `ErrorReturnCode` exception class [sets the exception message to a Unicode string](https://github.com/amoffat/sh/blob/858adf0c682af4c40e41f34d6926696b7a5d3b12/sh.py#L384-L388), regardless of Python version. That's not really an option in Python 2, but it just happens to work when your Unicode string is ASCII safe.

However, when the process exited with non-ASCII characters on stderr or stdout, then showing the exception in a traceback results in a `<exception str() failed>` exception message, because the `str()` call on the exception instance fails.

Demo reproducing the problem, run with Python 2.7:

```python
import sh, sys
print(sys.version_info)
python27 = sh.Command(sys.executable)
python27(c=r"import sys; sys.stderr.write(u'\u2603\n'.encode('utf8')); sys.exit(1)")
```

This outputs

```python
sys.version_info(major=2, minor=7, micro=15, releaselevel='final', serial=0)
  File "/.../lib/python2.7/site-packages/sh.py", line 1427, in __call__
  File "/.../lib/python2.7/site-packages/sh.py", line 774, in __init__
    self.wait()
  File "/.../lib/python2.7/site-packages/sh.py", line 792, in wait
  File "/.../lib/python2.7/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: <exception str() failed>
```

When you catch the exception, the `.args[0]` attribute shows why:

```python
>>> try:
...     python27(c=r"import sys; sys.stderr.write(u'\u2603\n'.encode('utf8')); sys.exit(1)")
... except sh.ErrorReturnCode as exc:
...     print(repr(exc.args[0]))
...     str(exc)
...
u"\n\n  RAN: /.../bin/python -c import sys; sys.stderr.write(u'\\u2603\\n'.encode('utf8')); sys.exit(1)\n\n  STDOUT:\n\n\n  STDERR:\n\u2603\n"
  File "<stdin>", line 5, in <module>
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2603' in position 129: ordinal not in range(128)
```

So on Python 2, please encode the message to a suitable `str` object before setting it.

abadger:
In Python3, the str type is a text string.  On Python2, the str type is
a byte string.  When we raise exceptions we should use native strings
otherwise Python may end up truncating or omitting the exception
message (on Python2) or mangling them (if a byte string were to be given
on Python3).

Fixes #463 

(Note, if you want some more examples of this outside of sh code, you can take a look at this documentation I wrote for the kitchen library: https://pythonhosted.org/kitchen/unicode-frustrations.html#frustration-5-exceptions )

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -387,6 +387,12 @@ class ErrorReturnCode(Exception):
             stderr=exc_stderr.decode(DEFAULT_ENCODING, "replace")
         )
 
+        if not IS_PY3:
+            # Exception messages should be treated as an API which takes native str type on both
+            # Python2 and Python3.  (Meaning, it's a byte string on Python2 and a text string on
+            # Python3)
+            msg = encode_to_py3bytes_or_py2str(msg)
+
         super(ErrorReturnCode, self).__init__(msg)
 
 
--- a/test.py
+++ b/test.py
@@ -245,16 +245,15 @@ class FunctionalTests(BaseTests):
         py = create_tmp_test("exit(1)")
 
         arg = "漢字"
+        native_arg = arg
         if not IS_PY3:
             arg = arg.decode("utf8")
+
 
         try:
             python(py.name, arg, _encoding="utf8")
         except ErrorReturnCode as e:
-            if IS_PY3:
-                self.assertTrue(arg in str(e))
-            else:
-                self.assertTrue(arg in unicode(e))
+            self.assertTrue(native_arg in str(e))
         else:
             self.fail("exception wasn't raised")
 


amoffat:
Gnarly...I had to really think about this one.  Thanks for the fix, it will go out in the next release :+1:

PR #469, opened 2019-03-24
fixed code-block-python: IndentationError by replace tabs with spaces

akgnah:
There is currently an IndentationError at http://amoffat.github.io/sh/tutorials/interacting_with_processes.html


amoffat:
:+1:

PR #472, opened 2019-04-20
Make ErrorReturnCode "RAN:" copy-pastable.

pteromys:
Using shlex.quote.

Before:
  sh.ErrorReturnCode_1:
    RAN: /usr/bin/sh -c echo -rf build && ls / && exit 1

After:
  sh.ErrorReturnCode_1:
    RAN: /usr/bin/sh -c 'echo -rf build && ls / && exit 1'

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -87,6 +87,11 @@ else:
     from io import BytesIO as iocStringIO
     from Queue import Queue, Empty
 
+try:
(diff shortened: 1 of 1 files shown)


amoffat:
Cool :+1: .  It will go out with the next release.

PR #482, opened 2019-07-03
Use raw string to avoid invalid escape deprecation warning.

thatch:
See https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior for this newly added warning as of 3.6.  This change should be fully backwards compatible.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -421,7 +421,7 @@ class CommandNotFound(AttributeError): pass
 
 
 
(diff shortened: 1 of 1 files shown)


thatch:
Failing tests are because of travis problems, not my change.

thatch:
Ping @amoffat ?

amoffat:
Looks correct @thatch, thank you for the contribution.  There's a lot of work involved in shipping the next release, and I'm very busy these days, so it may be awhile before it lands into master.

amoffat:
On `develop`, will go out soon.  Thanks again

PR #488, opened 2019-10-02
prefix True valued long options with the right prefix

FooBarQuaxx:


FooBarQuaxx:
It looks like the ci jobs are failing because the travis config requires non supported version of python (2.6 and 3.3) and pypy (5.3.1) for ubuntu xenial: https://docs.travis-ci.com/user/languages/python/#python-versions

amoffat:
Good work :+1:  Merged into `develop` and will go out with 1.14.15

Issue #460, opened 2018-11-15
wrong type for _env leads to confusing error


bulma-capsule:
When the type of **_env** is a **set** instead of a **dict** (a common error when using a coma instead of a colon), the error raised suggests the command itself failed and does not hint that the type of the **_env** argument is incorrect.

```python
>>> sh.env(_env={'FOO': 'bar'})
FOO=bar

>>> sh.env(_env={'FOO', 'bar'})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/venv/lib/python3.6/site-packages/sh.py", line 1427, in __call__
    return RunningCommand(cmd, call_args, stdin, stdout, stderr)
  File "/home/venv/lib/python3.6/site-packages/sh.py", line 774, in __init__
    self.wait()
  File "/home/venv/lib/python3.6/site-packages/sh.py", line 792, in wait
    self.handle_command_exit_code(exit_code)
  File "/home/venv/lib/python3.6/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_255: 

  RAN: /usr/bin/env

  STDOUT:


  STDERR:
```

singuliere:
ran into the same error today

ecederstrand:
Also hit this, but this time it was a bit more sneaky. All values in the env dict must be strings:

```python
>>> from sh import ls
>>> ls(_env={'MY_VALUE': "100"})
README.md
>>> ls(_env={'MY_VALUE': 100})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "lib/python3.7/site-packages/sh.py", line 1427, in __call__
    return RunningCommand(cmd, call_args, stdin, stdout, stderr)
  File "lib/python3.7/site-packages/sh.py", line 774, in __init__
    self.wait()
  File "lib/python3.7/site-packages/sh.py", line 792, in wait
    self.handle_command_exit_code(exit_code)
  File "lib/python3.7/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_255: 

  RAN: /bin/ls

  STDOUT:


  STDERR:

```

PR #491, opened 2019-11-25
Display exception information on 255 error

ssbarnea:
Fix issue where the origina exception information is lost when sh is
unable to write it the designated file descriptor.

This can happen with at least two identified reasons:
- passed cwd does not exist
- the output pipe is closed or invalid

In both cases the library failed without giving any meaningful message.

New behavior is to attempt to write exception information to sys.stderr,
which at least in my cases worked fine.


amoffat:
@ssbarnea cool.  Can you come up with a test case for this?  I want to be able to catch regressions if someone changes these lines.  Also, I believe (not 100% sure though), that "except as e" was introduced in a version of python after 2.4.  I am still supporting 2.4 for the time being (but I should probably rethink this).

I see the TravisCI checks are in a sad state right now due to broken Python support :(.  I will try to come up with a docker image suite for testing all applicable versions, but for now we can ignore Travis

amoffat:
@ssbarnea take a look at https://github.com/amoffat/sh/pull/493.  It should simplify running tests.  Want to give it a try for this PR?

ecederstrand:
I could have used this when I hit #460 

I have created the following test case (there's probably a less complicated way):

```python
    def test_255_exit_code_info(self):
        py = create_tmp_test("""
import sh
from sh import ls
# Monkey-patch the validator so we can pass an invalid
# env which triggers a 255 exit code
sh.special_kwarg_validator = lambda a, b: []
ls(_env='XXX')
""")
        with self.assertRaises(sh.ErrorReturnCode) as e:
            python(py.name, _err_to_out=True)
        self.assertIn(
            b"FATAL: 'str' object has no attribute 'keys'",
            e.exception.stdout
        )
```

This did require me to rewrite the original patch slightly to get improved formatting of the stack trace in the `FATAL` part. On Python3, sys.stderr expects unicode, so `tb` should also be included as unicode:

```python
            except Exception as e:
                # some helpful debugging
                tb = traceback.format_exc()
                try:
                    os.write(exc_pipe_write, tb.encode("utf8", "ignore"))
                except Exception as f:
                    # dump to stderr if we cannot save it to exc_pipe_write
                    if not exc_pipe_write or not hasattr(exc_pipe_write, 'write'):
                        sys.stderr.write(
                            "\nFATAL: %s\n%s" % (e, str(tb)))
                finally:
                    os._exit(255)
```

amoffat:
This 255 exception is fixed by 342f246dec4f4341cc65513286d3ecdcec152e5b and a4e2ef603b496e1c51f640ea55e9d76cf4a4b1ba.   You can read more about the underlying issue [here](https://github.com/amoffat/sh/pull/497#issuecomment-618831558).

I changed this PR a little bit because I like the idea of a fallback exception output, but I think the information that we report should be different.  As it was written, if the child couldn't write the traceback to the exception pipe, it wrote the traceback to sys.stderr.  However, if the child cannot write to the exception pipe, there are major system-level or critical sh flaws occurring, and so in that case, I think it is more relevant to report the reason why the exception pipe fd couldn't be written to, instead of the original traceback.

PR #492, opened 2019-11-25
Updated supported versions of Python

ssbarnea:
- dropped obsolete versions like py26, py34, py34
- added newer stabled releases


amoffat:
Merged into https://github.com/amoffat/sh/tree/develop.  Nice to see green tests again :) :+1:

PR #494, opened 2019-11-29
Fix for the fix for issue amoffat#398.

Cheaterman:
@davidm-mm proposed a fix for this issue in https://github.com/amoffat/sh/issues/398.
However it didn't work for me when using gevent inside docker (I suspect gevent is the issue, but have no evidence for/against it).
This seems to work for me.


amoffat:
:+1:

PR #497, opened 2020-01-10
Add a validator for env. Fixes #460

### LINKED ISSUE #460: wrong type for _env leads to confusing error

ecederstrand:
I had to disable old Python versions in Travis to get the test run to pass. These old versions can no longer be installed.


ecederstrand:
Test failures on Python 3.7 and 3.8 seem unrelated. I'll remove those from travis again and leave that for a different PR.

ecederstrand:
@amoffat Ping?

amoffat:
Great PR and tests :+1: 

The interesting thing is that you stumbled across a deeper problem.  There is a mechanism in place to catch fork-child exceptions and write the traceback out to the parent process (via a pipe).  The fact that we weren't seeing any output related to the env variable not being a dictionary means that that mechanism wasn't working correctly.  Digging deeper, it turns out that the pipe fd was being aggressively closed by the child process, as part of cleaning up inherited file descriptors.  342f246dec4f4341cc65513286d3ecdcec152e5b fixes that.  Now you get this nice exception:

```
======================================================================
ERROR: test_invalid_env (__main__.FunctionalTests)
  File "test.py", line 657, in test_invalid_env
    self.assertRaises(exc, ls, _env="XXX")
  File "/usr/lib/python3.5/unittest/case.py", line 733, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/usr/lib/python3.5/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)                   
  File "/home/amoffat/workspace/sh/sh.py", line 1457, in __call__
  File "/home/amoffat/workspace/sh/sh.py", line 767, in __init__
    self.call_args, pipe, process_assign_lock)                             
  File "/home/amoffat/workspace/sh/sh.py", line 2045, in __init__
    raise ForkException(fork_exc)    
sh.ForkException:                                      
                            
Original exception:                                             
===================      
                                                             
      File "/home/amoffat/workspace/sh/sh.py", line 2004, in __init__
        os.execve(cmd[0], cmd, ca["env"])                           
    AttributeError: 'str' object has no attribute 'keys'
```

I imagine this was a pain to debug going by only the 255 exit code.

amoffat:
Just to clarify, your validator commit will ensure that the above traceback snippet won't be seen (I had to disable your validator to get that output), and we'll get your friendly error text, but future 255 exit code errors will have the child exception visible at least.

ecederstrand:
Thanks for the explanation and extra fix!

It was indeed a pain to debug. I was chasing a timeout error in a painfully long test suite running inside a docker container. I was playing around with different timeout values for a certain part of the test, passed via `_env`, and only after several days it occurred to me that I was passing `_env={'TIMEOUT': 300}` and not `_env={'TIMEOUT': '300'}`. The test wasn't failing. `sh` was :-)

amoffat:
I am tempted to add some kind of environment sanitizer function that goes through all the key-values and stringifies them.  Can you think of any downsides?

ecederstrand:
Since the fix is very easy for the user (just add `str()` around your key or value), and the exception message should explain the error clearly, I think it may be better to just leave it as-is.

I can't think of any catastrophe's off the top of my head, but in some people use `__str__` for all kinds of weird tricks. Mistakenly passing a key or value that just happens to support `str()` could lead to hard-to-debug situations.

Bash does support 'integers' in env values (`export FOO=100` is the same as `export FOO="100"`) but since bash variables are actually untyped (see http://tldp.org/LDP/abs/html/untyped.html) you should should think of `100` as an unquoted string.

PR #525, opened 2020-05-03
None should not render an arg. closes #522

### LINKED ISSUE #522: Some way to easily handle boolean options
I've been writing some scripts lately that make heavy use of including/excluding options based on some boolean flag. It's clunky:

```python
if dry_run:
    sh.rsync("-n", ...)
else:
    sh.rsync(...)
```
It would be nicer if there was something, maybe `sh.Cond` that takes an object that evaluates to truthy, and a flag to be set if that evaluation is `True`.  So:

```python
sh.rsync(sh.Cond(dry_run, "-n"), ...)
```

I'm not sure if it should be a class (`Cond`) or a function (`cond`), or something named different altogether, but I would want to minimize its chances of colliding its name with other binaries on the system. Thoughts? @ecederstrand

amoffat:
Pretty simple. Before, a `None` argument passed to a command was being passed in as a string. Now, it omits the argument altogether.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -1581,7 +1581,7 @@ def compile_args(args, kwargs, sep, prefix):
                 processed_args.append(encode(sub_arg))
         elif isinstance(arg, dict):
             processed_args += aggregate_keywords(arg, sep, prefix, raw=True)
-        elif arg is None:
+        elif arg is None or arg is False:
             pass
         else:
             processed_args.append(encode(arg))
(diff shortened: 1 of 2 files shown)


amoffat:
> Can we just check if the value is falsy? I assume we can also safely ignore `''` in args.

I think there are cases where an empty string should be allowed in an argument, for example some program where you're replacing all occurrences of a string with empty string.

ecederstrand:
Good point.

What do you think about validating args to be strings after filtrering away `None` and `False`?

PR #530, opened 2020-05-12
_env just needs to be a Mapping

### LINKED ISSUE #527: `_env` param does not work with dict-like objects
https://github.com/amoffat/sh/blob/114e6eedf5db022d2a0e5a5b5104ea18394da1a6/sh.py#L1149-L1151
wrt the above snippet, is it necessary to enforce `dict`? This breaks compatibility with other dict-like objects. eg: `os.environ` returns a dict-like object `os._Environ`.

Perhaps a check for `isinstance(env, collections.Mapping)` would allow for more general interface?

ecederstrand:
This allows other values, as long as they implement the Mapping interface, i.e. `.items()`.

Fixes #527

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -52,7 +52,7 @@ import inspect
 import tempfile
 import warnings
 import stat
-from collections import deque
+from collections import deque, Mapping
 import glob as glob_module
 import ast
 from contextlib import contextmanager
@@ -110,7 +110,6 @@ import tty
 import fcntl
 import struct
 import resource
-from collections import deque
 import logging
 import weakref
 
@@ -1146,8 +1145,8 @@ def env_validator(passed_kwargs, merged_kwargs):
     if env is None:
         return invalid
 
-    if not isinstance(env, dict):
-        invalid.append(("env", "env must be a dict. Got {!r}".format(env)))
+    if not isinstance(env, Mapping):
+        invalid.append(("env", "env must be dict-like. Got {!r}".format(env)))
         return invalid
 
     for k, v in passed_kwargs["env"].items():
--- a/test.py
+++ b/test.py
@@ -558,6 +558,11 @@ print(dict(HERP=sh.HERP))
         out = python(py.name, _env=env, _cwd=THIS_DIR).strip()
         self.assertEqual(out, "{'HERP': 'DERP'}")
 
+        # Test that _env also accepts os.environ which is a mpping but not a dict.
+        os.environ["HERP"] = "DERP"
+        out = python(py.name, _env=os.environ, _cwd=THIS_DIR).strip()
+        self.assertEqual(out, "{'HERP': 'DERP'}")
+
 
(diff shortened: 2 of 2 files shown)


saytosid:
Hey sorry. I understand this is landed, I just had a minor comment.

The error message seems misleading - 
https://github.com/amoffat/sh/blob/fa4b1e370f2d4e2dd1e6183de8f3311fe410af23/sh.py#L1148-L1150

There can be an object which implements the `Mapping Protocol` completely from scratch without using `collections.Mapping`. In this case the object would be `dict-like` but not an `instanceof(_, collections.Mapping)` 🤔 

Perhaps it could be rephrased to convey the need explicitly?

amoffat:
Good point @saytosid. If we wanted to be really thorough, we would check what methods `os.spawnve` calls on the env dict in order to test for those directly. Otherwise it is difficult to know if a custom dict-like object implements enough of the correct methods to be acceptable.

https://github.com/amoffat/sh/blob/fa4b1e370f2d4e2dd1e6183de8f3311fe410af23/sh.py#L1520

Either that or we only allow `_env` to be an instance of `collections.Mapping` but to your point, that seems restrictive.

saytosid:
Indeed. But, my point was about the error message. I'm sorry I wasn't very clear.

`"env must be dict-like. Got {!r}"` - does not convey the restriction clearly. Maybe something like - `env must be collections.Mapping ...`?

ecederstrand:
I did think about the wording of the error message, and went for an intentionally vague text.

For a seasoned Python developer, you would probably know what `collections.Mapping` is and know what to do about the error message, or be good enough with a debugger to find the actual assertion code if you are doing advanced things.

For a more junior developer, I'm concerned that mentioning `collections.Mapping` in the error message is more confusing than helpful. They certainly wouldn't know that `dict` implements that interface, and I think in this case pointing in the direction of the collections module is unhelpful. In 99% of the error situations I can envision, you just want to pass a `dict` instead of whatever object was mistakenly passed.

I'm not fond of requiring less than `collections.Mapping`. You never know what the underlying `os.spawnve` implementation decides to do with the env. Not everyone will be running CPython.

I'm prepared to be convinced otherwise in both cases, of course.

saytosid:
>For a more junior developer, I'm concerned that mentioning collections.Mapping in the error message is more confusing than helpful. They certainly wouldn't know that dict implements that interface, and I think in this case pointing in the direction of the collections module is unhelpful. In 99% of the error situations I can envision, you just want to pass a dict instead of whatever object was mistakenly passed.

Sounds reasonable to me 👍 

> I'm not fond of requiring less than collections.Mapping. You never know what the underlying os.spawnve implementation decides to do with the env. Not everyone will be running CPython.

Definitely 💯

amoffat:
Sounds good to me @ecederstrand :+1:

PR #531, opened 2020-05-27
Use id() for equality between running commands

### LINKED ISSUE #518: Does RunningCommand.__eq__ make sense?
It was the root cause of https://github.com/amoffat/sh/issues/413

Here is [`RunningCommand.__eq__`](https://github.com/amoffat/sh/blob/develop/sh.py#L923). Basically the issue was that attempting to call `list.remove(p)` on a list of processes was forcing many of those processes to evaluate/finish, due to `__eq__` indirectly calling `wait()`.

The question is, does it even make sense to compare processes like this? You might be able to have `__eq__` just compare pids, but pids can also be re-used so it isn't perfect.

@ecederstrand what do you think?

ecederstrand:
Use id() instead of the expensive unicode() that has the side effect of actually running the command if it is not already running, and blocking until the command has finished running.

Closes #518

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
+
+## 2.0.0 - X/XX/XX
+*   `RunningCommand.__eq__` no longer has the side effect of executing the command [#518](https://github.com/amoffat/sh/pull/531)
+
 ## 1.14.0 - X/XX/XX
 *   `_env` now more lenient in accepting dictionary-like objects [#527](https://github.com/amoffat/sh/issues/527)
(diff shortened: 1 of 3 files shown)


amoffat:
I didn't realize so many tests would fail from this, but it makes sense. I suppose wrapping all of the commands with `str()` before comparing their output would fix them.

amoffat:
Good work. The tests breaking made me realize just how much this changes existing behavior, so I went ahead and changed the branch base to be `2.0.0`. We can put the things that break compatibility in that branch.

ecederstrand:
Good idea to push it to `2.0.0`. Branch should be ready for merge.

PR #532, opened 2020-05-28
File cleanup

ecederstrand:
Sweeping cleanup for:
* Trailing whitespace
* Imprecise assert methods
* Linter warnings

@amoffat I'm thinking of also fixing linter warnings in sh.py and test.py, just because my editor has a huge number of inspection warnings. Would you mind running e.g. flake8 as part of the test suite?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,6 @@
 [run]
 branch = True
-source = 
+source =
     sh.py
 
 [report]
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ before_script:
 
 script:
     - python sh.py travis
+    - python -m flake8 sh.py
     - if python -c 'import sys; sys.exit(int(not sys.version_info >= (3, 5)))' ; then python setup.py check --restructuredtext --metadata --strict ; fi
 
 after_success:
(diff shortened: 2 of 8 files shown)


amoffat:
Flake8 sounds like a great idea @ecederstrand

ecederstrand:
Ready for review.

There were a bunch of warnings on re-using variable names from outer scope. Most were easy to rename, but one was painful: there's a top-level `sh.args` that is part of the API. This means I had to rename `args` in all `def foo(*args, **kwargs)` definitions even though that's a convention. Ugh. I renamed most to just `def foo(*a, **kwargs)` in lack of a better suggestion.

ecederstrand:
Thanks for fixing the test failures!

The warning that caused me to do the `args->a` rename is actually useful enough; it warns about overwriting variables from outer scope. Consider:

```python
@contextmanager
def args(**kwargs):
    # This context manager is part of the API
    pass

def foo(*args, **kwargs):
    # Now we can't access the outer args()
    pass
```

In this particular case, I think it may be overkill to fix, unless we can use some import magic to rename `def args()` to e.g. `def _args()` and have `from sh import args` still work.

ecederstrand:
Since `sh.args` is deprecated anyway, I took the liberty of renaming `sh.args()` to `sh._args()` and adding a hack in `Environment` so it's still importable as `sh.args`· That allows to revert the ugly `*args -> *a` rename.

PR #536, opened 2020-08-27
Adding find_spec to ModuleImporterFromVariables

sbacchio:
This PR adds the method `find_spec` to `ModuleImporterFromVariables`.

The method `find_module` (implemented in `ModuleImporterFromVariables`) has been deprecated since Python 3.4 in favor of `find_spec`: https://docs.python.org/3/library/sys.html#sys.meta_path

The absence of `find_spec` made me incur in an issue with pytest: https://github.com/pytest-dev/pytest/issues/7689.
This PR would resolve that issue.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -22,7 +22,7 @@ http://amoffat.github.io/sh/
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 # ===============================================================================
-__version__ = "1.13.2"
+__version__ = "1.13.1"
 __project_url__ = "https://github.com/amoffat/sh"
 
 from collections import deque, Mapping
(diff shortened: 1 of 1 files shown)


sh.py
```diff
@@ -22,7 +22,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 # ===============================================================================
-__version__ = "1.13.1"
+__version__ = "1.13.2"
```
ecederstrand, line 25: Please don't bump the package version here. We'll do that in a separate commit when we are ready to release a new version.

sh.py
```diff
@@ -3651,6 +3654,13 @@ def find_module(self, mod_fullname, path=None):
 
         return self
 
+    def find_spec(self, fullname, path=None, target=None):
```
ecederstrand, line 3657: Please change the string comment to proper docsting format (triple-quote).

We normally don't do lazy imports, but I don't see a good way around it because we also don't do top-level imports for things that don't have an equivalent in older Python versions. Let's leave it here.

We could consider putting the `find_spec()` definition itself under `if IS_PY3 and MINOR_VER >= 4: def find_spec(...` but as long as we have the comment, I don't really see the value.

ecederstrand (changes requested):
Thanks for the contribution!

sh.py
sbacchio: Done

sh.py
sbacchio: Ok, sorry about that!

amoffat:
:+1: Good work all. Release 1.14.0 is out with these changes

PR #537, opened 2020-09-01
Fallback import for collections.abc.Mapping

dvzrv:
sh.py:
Since python 3.3 the abstract classes in the collections module are
ordered below collections.abc.
While until python 3.8 this only shows a warning during import it will
be raising an ImportError with python 3.9.
Therefore the Mapping class is imported from collections.abc by default
and falls back to importing from collections directly for python < 3.3
when encountering an ImportError.


ecederstrand:
Thanks!

PR #540, opened 2020-09-01
Remove hardcoded path for sleep executable

### LINKED ISSUE #539: test_timeout raises AssertionError
Hi! When packaging 1.14.0 for Arch Linux I ran the tests using pytest (i.e. `pytest -v test.py`).
Unfortunately the location of the `sleep` executable is hardcoded to be `/bin/sleep` whereas on Arch Linux it is in `/usr/bin/sleep`:

```
=================================== FAILURES ===================================
_________________________ FunctionalTests.test_timeout _________________________

self = <test.FunctionalTests testMethod=test_timeout>

    def test_timeout(self):
        import sh
        from time import time

        sleep_for = 3
        timeout = 1
        started = time()
        try:
>           sh.sleep(sleep_for, _timeout=timeout).wait()

/build/python-sh/src/sh-1.14.0/test.py:2015:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Command '/usr/bin/sleep'>, args = [3], kwargs = {}
cmd = [b'/usr/bin/sleep', b'3']
call_args = {'arg_preprocess': None, 'bg': False, 'bg_exc': True, 'close_fds': True, ...}
preprocessor = None, extracted_call_args = {'timeout': 1}, stdin = None
first_arg = 3, processed_args = [b'3'], split_args = [b'3']

    def __call__(self, *args, **kwargs):
        kwargs = kwargs.copy()
        args = list(args)

        # this will hold our final command, including arguments, that will be
        # exec'ed
        cmd = []

        # this will hold a complete mapping of all our special keyword arguments
        # and their values
        call_args = Command._call_args.copy()

        # aggregate any 'with' contexts
        for prepend in get_prepend_stack():
            pcall_args = prepend.call_args.copy()
            # don't pass the 'with' call arg
            pcall_args.pop("with", None)

            call_args.update(pcall_args)
            cmd.extend(prepend.cmd)

        cmd.append(self._path)

        # do we have an argument pre-processor?  if so, run it.  we need to do
        # this early, so that args, kwargs are accurate
        preprocessor = self._partial_call_args.get("arg_preprocess", None)
        if preprocessor:
            args, kwargs = preprocessor(args, kwargs)

        # here we extract the special kwargs and override any
        # special kwargs from the possibly baked command
        extracted_call_args, kwargs = self._extract_call_args(kwargs)

        call_args.update(self._partial_call_args)
        call_args.update(extracted_call_args)

        # handle a None.  this is added back only to not break the api in the
        # 1.* version.  TODO remove this in 2.0, as "ok_code", if specified,
        # should always be a definitive value or list of values, and None is
        # ambiguous
        if call_args["ok_code"] is None:
            call_args["ok_code"] = 0

        if not getattr(call_args["ok_code"], "__iter__", None):
            call_args["ok_code"] = [call_args["ok_code"]]

        # check if we're piping via composition
        stdin = call_args["in"]
        if args:
            first_arg = args.pop(0)
            if isinstance(first_arg, RunningCommand):
                if first_arg.call_args["piped"]:
                    stdin = first_arg.process
                else:
                    stdin = first_arg.process._pipe_queue

            else:
                args.insert(0, first_arg)

        processed_args = compile_args(args, kwargs, call_args["long_sep"], call_args["long_prefix"])

        # makes sure our arguments are broken up correctly
        split_args = self._partial_baked_args + processed_args

        final_args = split_args

        cmd.extend(final_args)

        # if we're running in foreground mode, we need to completely bypass
        # launching a RunningCommand and OProc and just do a spawn
        if call_args["fg"]:

            cwd = call_args["cwd"] or os.getcwd()
            with pushd(cwd):
                if call_args["env"] is None:
                    exit_code = os.spawnv(os.P_WAIT, cmd[0], cmd)
                else:
                    exit_code = os.spawnve(os.P_WAIT, cmd[0], cmd, call_args["env"])

            exc_class = get_exc_exit_code_would_raise(exit_code, call_args["ok_code"], call_args["piped"])
            if exc_class:
                if IS_PY3:
                    ran = " ".join([arg.decode(DEFAULT_ENCODING, "ignore") for arg in cmd])
                else:
                    ran = " ".join(cmd)
                exc = exc_class(ran, b"", b"", call_args["truncate_exc"])
                raise exc
            return None

        # stdout redirection
        stdout = call_args["out"]
        if output_redirect_is_filename(stdout):
            stdout = open(str(stdout), "wb")

        # stderr redirection
        stderr = call_args["err"]
        if output_redirect_is_filename(stderr):
            stderr = open(str(stderr), "wb")

>       return RunningCommand(cmd, call_args, stdin, stdout, stderr)

/build/python-sh/src/sh-1.14.0/sh.py:1520:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = , cmd = [b'/usr/bin/sleep', b'3']
call_args = {'arg_preprocess': None, 'bg': False, 'bg_exc': True, 'close_fds': True, ...}
stdin = None, stdout = None, stderr = None

    def __init__(self, cmd, call_args, stdin, stdout, stderr):
        """
            cmd is a list, where each element is encoded as bytes (PY3) or str (PY2)
        """

        # self.ran is used for auditing what actually ran.  for example, in
        # exceptions, or if you just want to know what was ran after the
        # command ran
        #
        # here we're making a consistent unicode string out if our cmd.
        # we're also assuming (correctly, i think) that the command and its
        # arguments are the encoding we pass into _encoding, which falls back to
        # the system's encoding
        enc = call_args["encoding"]
        self.ran = " ".join([shlex_quote(arg.decode(enc, "ignore")) for arg in cmd])

        self.call_args = call_args
        self.cmd = cmd

        self.process = None
        self._waited_until_completion = False
        should_wait = True
        spawn_process = True

        # this is used to track if we've already raised StopIteration, and if we
        # have, raise it immediately again if the user tries to call next() on
        # us.  https://github.com/amoffat/sh/issues/273
        self._stopped_iteration = False

        # with contexts shouldn't run at all yet, they prepend
        # to every command in the context
        if call_args["with"]:
            spawn_process = False
            get_prepend_stack().append(self)

        if call_args["piped"] or call_args["iter"] or call_args["iter_noblock"]:
            should_wait = False

        # we're running in the background, return self and let us lazily
        # evaluate
        if call_args["bg"]:
            should_wait = False

        # redirection
        if call_args["err_to_out"]:
            stderr = OProc.STDOUT

        done_callback = call_args["done"]
        if done_callback:
            call_args["done"] = partial(done_callback, self)

        # set up which stream should write to the pipe
        # TODO, make pipe None by default and limit the size of the Queue
        # in oproc.OProc
        pipe = OProc.STDOUT
        if call_args["iter"] == "out" or call_args["iter"] is True:
            pipe = OProc.STDOUT
        elif call_args["iter"] == "err":
            pipe = OProc.STDERR

        if call_args["iter_noblock"] == "out" or call_args["iter_noblock"] is True:
            pipe = OProc.STDOUT
        elif call_args["iter_noblock"] == "err":
            pipe = OProc.STDERR

        # there's currently only one case where we wouldn't spawn a child
        # process, and that's if we're using a with-context with our command
        self._spawned_and_waited = False
        if spawn_process:
            log_str_factory = call_args["log_msg"] or default_logger_str
            logger_str = log_str_factory(self.ran, call_args)
            self.log = Logger("command", logger_str)

            self.log.debug("starting process")

            if should_wait:
                self._spawned_and_waited = True

            # this lock is needed because of a race condition where a background
            # thread, created in the OProc constructor, may try to access
            # self.process, but it has not been assigned yet
            process_assign_lock = threading.Lock()
            with process_assign_lock:
                self.process = OProc(self, self.log, cmd, stdin, stdout, stderr,
                                     self.call_args, pipe, process_assign_lock)

            logger_str = log_str_factory(self.ran, call_args, self.process.pid)
            self.log.context = self.log.sanitize_context(logger_str)
            self.log.info("process started")

            if should_wait:
>               self.wait()

/build/python-sh/src/sh-1.14.0/sh.py:784:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = , timeout = None

    def wait(self, timeout=None):
        """ waits for the running command to finish.  this is called on all
        running commands, eventually, except for ones that run in the background

        if timeout is a number, it is the number of seconds to wait for the process to resolve. otherwise block on wait.

        this function can raise a TimeoutException, either because of a `_timeout` on the command itself as it was
        launched, or because of a timeout passed into this method.
        """
        if not self._waited_until_completion:

            # if we've been given a timeout, we need to poll is_alive()
            if timeout is not None:
                waited_for = 0
                sleep_amt = 0.1
                alive = False
                exit_code = None
                if timeout < 0:
                    raise RuntimeError("timeout cannot be negative")

                # while we still have time to wait, run this loop
                # notice that alive and exit_code are only defined in this loop, but the loop is also guaranteed to run,
                # defining them, given the constraints that timeout is non-negative
                while waited_for <= timeout:
                    alive, exit_code = self.process.is_alive()

                    # if we're alive, we need to wait some more, but let's sleep before we poll again
                    if alive:
                        time.sleep(sleep_amt)
                        waited_for += sleep_amt

                    # but if we're not alive, we're done waiting
                    else:
                        break

                # if we've made it this far, and we're still alive, then it means we timed out waiting
                if alive:
                    raise TimeoutException(None, self.ran)

                # if we didn't time out, we fall through and let the rest of the code handle exit_code.
                # notice that we set _waited_until_completion here, only if we didn't time out. this allows us to
                # re-wait again on timeout, if we catch the TimeoutException in the parent frame
                self._waited_until_completion = True

            else:
                exit_code = self.process.wait()
                self._waited_until_completion = True

            if self.process.timed_out:
                # if we timed out, our exit code represents a signal, which is
                # negative, so let's make it positive to store in our
                # TimeoutException
>               raise TimeoutException(-exit_code, self.ran)
E               sh.TimeoutException
/build/python-sh/src/sh-1.14.0/sh.py:838: TimeoutException

During handling of the above exception, another exception occurred:

self = <test.FunctionalTests testMethod=test_timeout>

    def test_timeout(self):
        import sh
        from time import time

        sleep_for = 3
        timeout = 1
        started = time()
        try:
            sh.sleep(sleep_for, _timeout=timeout).wait()
        except sh.TimeoutException as e:
>           self.assertEqual(e.full_cmd, '/bin/sleep 3')
E           AssertionError: '/usr/bin/sleep 3' != '/bin/sleep 3'
E           - /usr/bin/sleep 3
E           ? ----
E           + /bin/sleep 3

/build/python-sh/src/sh-1.14.0/test.py:2017: AssertionError
=========================== short test summary info ============================
FAILED test.py::FunctionalTests::test_timeout - AssertionError: '/usr/bin/sle...
=========== 1 failed, 165 passed, 1 skipped, 3 deselected in 45.41s ============
```

This error can be circumvented by changing the respective line of code, but I think using e.g. `"sleep 3" in e.full_cmd` in the assertion would fix this issue more generically.

dvzrv:
test.py:
In test_timeout the path for the sleep executable has been hardcoded to
be /bin/sleep. However, on operating systems such as Arch Linux, Fedora
or Solaris the executable resides in /usr/bin/sleep due to a /usr merge
(e.g. see
  https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/).
By only checking for the name of the executable and its parameter the
test becomes more generic and thus can run on any (Unix-like) operating
system (given that `sleep` is in PATH).

Fixes #539

### DIFF
--- a/test.py
+++ b/test.py
@@ -2014,7 +2014,7 @@ sys.stdout.write("line1")
         try:
             sh.sleep(sleep_for, _timeout=timeout).wait()
         except sh.TimeoutException as e:
-            self.assertEqual(e.full_cmd, '/bin/sleep 3')
+            assert 'sleep 3' in e.full_cmd
         else:
             self.fail("no timeout exception")
         elapsed = time() - started


amoffat:
Looks good, thanks

PR #555, opened 2021-02-10
Cast argument to pushd as str to support passing Paths

MorganRodgers:
Support using [Path](https://docs.python.org/3/library/pathlib.html) as the argument to `sh.pushd` by explicitly casting to `str`.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -3191,7 +3191,7 @@ def pushd(path):
     context, unlike the _cwd arg this will work with other built-ins such as
     sh.glob correctly """
     orig_path = os.getcwd()
-    os.chdir(path)
+    os.chdir(str(path))
     try:
         yield
     finally:


ecederstrand:
@amoffat The diff itself looks good to me, but it does open up the discussion whether `sh` should accept `Path` objects more generally.

`os` primitives started to accept `Path` objects (see https://docs.python.org/3/library/os.html#os.chdir) from Python 3.6. All earlier versions of Python are EOL, so this is only an issue for `sh` on old Python versions.

Supporting `Path` objects in `pushd` may mislead users to think that `sh` also supports Path objects elsewhere. And I'm not convinced that ensuring that is worthwhile.

MorganRodgers:
When used with 3.6 and up I'm not aware of any instances where `sh` doesn't gracefully handle `Paths`. In general I've been very pleased with how `sh` can accept Paths when they're passed as arguments to subprocesses. `pushd` was a stand out to me since it failed, but then it has been a while since I've had to work on a 2.7 project.

I'm fine with closing this since 2.x is EOL.

amoffat:
Great background @ecederstrand and I agree with your conclusions. @MorganRodgers great, I will close then.

PR #569, opened 2021-05-16
Release/v2.0.0

amoffat:


sh.py
```diff
@@ -366,39 +309,38 @@ def __reduce__(self):
         return self.__class__, (self.full_cmd, self.stdout, self.stderr, self.truncate)
 
     def __init__(self, full_cmd, stdout, stderr, truncate=True):
+        self.exit_code = self.exit_code  # makes pylint happy
```
honnix, line 312: This doesn't seem to work. From a fresh installation of 2.0.3:

```python
>>> from sh import ErrorReturnCode
>>> ErrorReturnCode("", None, None)
  File ".../site-packages/sh.py", line 313, in __init__
    self.exit_code = self.exit_code  # makes pylint happy
AttributeError: 'ErrorReturnCode' object has no attribute 'exit_code'
```

ecederstrand: `ErrorReturnCode` instances are not meant to be created like that. They are created from the metaclass. See e.g. https://github.com/amoffat/sh/blob/develop/sh.py#L445

If you have an actual use case for the above example code, then please open an issue and we can discuss there.

PR #570, opened 2021-05-22
remove thread toplevel reraise to not pollute curses screen

saulpw:


saulpw:
Hi @amoffat, thanks for the `sh` library!  VisiData uses it for some of its plugins, and we had some screen corruption when errors happened.  This fixed the issue for me and I've been maintaining a tiny fork, but if you're willing to take this PR then VisiData can depend on your official and most up-to-date `sh` repository again.  Either way, feel free to edit or close the PR as needed.  Thanks!

ecederstrand:
Hi @saulpw Thanks for the contribution!

Can you provide an example showing the corruption so we can better understand and possibly reproduce the issue?

Also, I don't understand the symlink part of the PR. Can you please explain what that solves?

saulpw:
Sorry about that, I was using `sh` as a submodule and it was a convenient way to import it, but I didn't realize it was at your toplevel.  I'm reorganizing anyway and it should be a proper dependency instead of a submodule.  Removed that commit.

As for the other one, we were using sh to run commands in the background, and when those commands exited with errors, the re-raise is a toplevel exception which Python unavoidably sends to stderr, thereby corrupting the screen output.  Seems like this would only be needed in an interactive REPL, otherwise the process and its exit code would need to be captured separately anyway.

ecederstrand:
Let me post some examples to try to understand your problem. Here's an example where a process returns a non-zero exit code. There's no issue when you run it in the foreground because you can just catch the exception and handle reporting as you want:

```python
>>> from sh import bash
>>> p = bash(c='sleep 1; exit 1', _bg=False)
[...]
sh.ErrorReturnCode_1: 

  RAN: /bin/bash -c 'sleep 1; exit 1'

  STDOUT:

  STDERR:
```

Now you run the same, but in the background. You get the same stack trace, but you can't catch it if you don't want to see it because the statement returns immediately.
```python

>>> p = bash(c='sleep 1; exit 1', _bg=True)
>>>
Exception in thread background thread for pid 83723:
[...]
sh.ErrorReturnCode_1: 

  RAN: /bin/bash -c 'sleep 1; exit 1'

  STDOUT:

  STDERR:
```

If this is indeed your issue, then you can use the special [_bg_exc](https://amoffat.github.io/sh/sections/special_arguments.html?highlight=bg_exc#bg-exc) kwarg. This allows you to hide the stack trace by ignoring it. You can still get the exception if you then call `.wait()`:

```python
>>> p = bash(c='sleep 1; exit 1', _bg=True, _bg_exc=False)
>>> # No stack trace!
>>> p.wait()
[...]
sh.ErrorReturnCode_1: 

  RAN: /bin/bash -c 'sleep 1; exit 1'

  STDOUT:

  STDERR:
```

saulpw:
Thanks for the detailed explanation, @ecederstrand!  Looks like `_bg_exc` was the option I needed.

Issue #559, opened 2021-03-03
Execution context (default arguments) for custom command?


jerkstorecaller:
I want sh to output to always output stdout/stderr as if I'd run the commands in a bash script, for every single command. Using the doc at https://amoffat.github.io/sh/sections/default_arguments.html as guide, I created a new sh2 object with my default argument for _out. However, it doesn't work for custom Commands, only for built-in commands.

```
import sh
import sys

sh2 = sh(_out=sys.stdout, _err=sys.stderr)

#correctly prints the output of "ls /etc" to stdout
sh2.ls("/etc")

subgit = sh2.Command("tools/subgit-3.3.11/bin/subgit")
#prints nothing
subgit("--version")
#prints the version correctly
subgit("--version", _out=sys.stdout)
```

It's not specific to this tool, it's the same result if I use ```myls = sh2.Command("/usr/bin/ls")``` .

amoffat:
Huh, I think you have found a bug, and I think it is reasonable to expect that `Command` instances should inherit the default arguments from `sh2`. Thanks for reporting

ecederstrand:
I can't reproduce this on my own machine.

@amoffat  The beginning of `SelfWrapper.__call__` looks very innocent. You've added a warning about some fragile code in https://github.com/amoffat/sh/blob/b6cb96ef920e720fa024a123df5365eebd8e463a/sh.py#L3527 and forward. I assume this is where things are going wrong. Do you have any ideas how to improve?

amoffat:
@ecederstrand here is what I am using to reproduce:

```python
import sh

sh2 = sh(_ok_code=1)

sh2.ls() # raises
ls = sh2.Command("ls")
ls() # doesn't raise
```

It seems that the arguments that should be "baked" into `sh2` aren't propagating to instances of the `Command` class

ecederstrand:
I created a minimal PR for this. The caveat is that `sh.Command("ls")` now also gets the `_ok_code=1` call arg. That's one step forward and one step back, IMHO. I can't see how to fix that without also creating wrapper code for `Command`. Any ides?

ecederstrand:
I've updated the PR without the above caveat and added a test case. Not entirely happy with the code - it seems there should be a less complicated way. Suggestions welcome :-)

amoffat:
LGTM @ecederstrand. Scoping the baked args to the environment-specific `Command` class is exactly how I would have tried to do it. Good work 👍

PR #573, opened 2021-06-04
Let Command know about baked call args when wrapping the module

ecederstrand:
Refs #559


amoffat (approved):
I think this is solid. I see what you mean about wanting a less complicated way, but I don't think that there is. It probably feels weird because you're messing with a lot of `_`-prefixed attributes, but I think it's ok, since it is all hidden from the user.

amoffat:
Would you also add a line to the changelog for this fix please? You can make a section for either `1.15.0` or `1.14.3` depending on what else we want to include in the release.

PR #584, opened 2021-08-28
Make sh.which() an internal function

### LINKED ISSUE #579: sh.which doesn't return an object and doesn't throw exceptions
Hi,
I experience this unexpected behavior:
Is this a bug, or am I doing something wrong?

Python Version:
```bash 
$ python --version
Python 3.9.5
```
sh Version: 1.14.2

Code snippet:

```python
 try:
      # i expect behavior as described here:
      # https://amoffat.github.io/sh/sections/exit_codes.html
      output = sh.which('xxxxx')
      # output returns a string or None , not an object
      # print(output.exit_code) fails because neither 'str' nor 'None'  has the attribute 'exit_code' 
 except ErrorReturnCode:
      # this exception is not thrown 
      logger.error("program not found")
```

ecederstrand:
Rename sh.which so normal 'from sh import which' hits the system binary.

Fixes #579


sh.py
```diff
@@ -3328,7 +3328,7 @@ def b_cd(path=None):
 
     @staticmethod
     def b_which(program, paths=None):
-        return which(program, paths)
+        return _which(program, paths)
```
amoffat, line 3331: The way that this is currently, I believe `sh.which` will continue to use `_which`, as it [supersedes any system commands](https://github.com/amoffat/sh/blob/develop/sh.py#L3295-L3303). But given that `which` is a shell builtin in `zsh` but in bash it relies on the system binary, I think we want to keep this `b_which` and just swap the checks around first, so that we first check for a system binary, and if it doesn't exist, we fall back to `b_which`.

ecederstrand: Thanks for the comment! I've changed the priority so binaries supersede built-ins. Even though zsh users may be surprised to now hit `/usr/bin/which`, I think it's fine because they were previously hitting our implementation of `which`, not the built-in from their shell.

This also changed `from sh import cd` because our built-in `cd` actually changes the CWD of the running Python process, while `fro sh import cd` now just does that in the subprocess, leaving the CWD of th parent process intact. I don't think we're worse off here, either.

amoffat: Good catch on the `cd` change. Thinking about it more, I think that might be breaking behavior. If `cd` doesn't change the directory of the python process, there isn't much use in it, since it can never be chained together with other commands. I am imagining that the only people who are relying on `cd` are relying on it because it changes the directory of all subsequent sh commands.

Maybe we should revert the 2622720131a1837cd9fca14b1cadfdf21fc68152 and instead make `b_which` test for the existence of a system `which`, and if it exists, use that, but if it doesn't, fall back to `_which`. Thoughts?

ecederstrand: As I see it, `sh` is just a wrapper around binaries, and executing a binary shouldn't change the state of the parent process. The fact that executing `cd()` doesn't do anything meaningful isn't really our problem. We can probably find other examples of that. If we didn't advertise the current behavior of `cd` in the docs, then people are just relying on unintended behavior.

On the other hand, I just googled `python cd context manager` because that's something I've needed before. And the first hit is a SO post talking about `sh`, with the person expecting to be able to use `cd` as a context manager. Which tells me 2 things:

1. The current behavior of `cd` is probably relied on, as you say, even though it violates the design of `sh`
2. It would be really cool if `cd` actually *was* a context manager

So, while current behavior is arguably a bug, nobody reported it before. Changing the current behavior would probably annoy lots of people for very little reason. I like your suggestion.

amoffat: I agree with your reasoning about `cd`.

Just to make things even more complicated, I just remembered that there is a [sh.pushd](https://amoffat.github.io/sh/sections/command_class.html?highlight=pushd#pushd) helper which behaves as a context manager. It behaves like a scoped `cd`. I think this is what we want to encourage people to use in the future.

So maybe we do this:
* proceed with preserving the original `cd` behavior for now (but fix the `which` behavior)
* In sh v2, we remove the `b_cd` wrapper and encourage people to use `pushd`

ecederstrand: I think that's a good plan. I had also forgot about `pushd` and I agree we should encourage using that.

ecederstrand: I've updated the PR to let internal `cd()` take precedence, and to let `sh.cd()` be used as a context manager.

PR #594, opened 2021-12-02
Added password kwarg

qbilius:
The example of using `sh.contrib.sudo` with a string password highly confused me as it was missing the crucial bit that the password must be passed via a keyword argument.


amoffat:
Thanks @qbilius, I can see how it can be misleading!

PR #598, opened 2022-02-05
Accept pathlib.Path objects for stdout/stderr

### LINKED ISSUE #597: Accept pathlib.Path objects for _out and _err
Currently, if I pass `pathlib.Path` object to any command as its `_out` argument, it is effectively ignored without any error:
```
import sh
from pathlib import Path

out = Path('/tmp/out')
sh.echo('hello', _out=out)
print(out.exists())  # False
```

Current workaround is to convert path to `string` before passing it to sh:
```
sh.echo('hello', _out=str(out))
print(out.exists())  # True
```

It would be nice to also accept Paths "as is", or at least raise an error if something unsupported is passed for `_out` or `_err`.

MarSoft:
Fix #597.
I decided to not add an additional import for `pathlib.Path` and instead check if the object passed is "path-like" according to [PEP-519](https://www.python.org/dev/peps/pep-0519/).

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -990,7 +990,7 @@ class RunningCommand(object):
 
 
 def output_redirect_is_filename(out):
-    return isinstance(out, basestring)
+    return isinstance(out, basestring) or hasattr(out, '__fspath__')
 
 
 def get_prepend_stack():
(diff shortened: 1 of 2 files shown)


amoffat:
Thanks for documenting the issue and starting a PR. Would you mind adding some tests around the usage of a `Path` object please?

mentalisttraceur:
Code review not: the Pythonic way to resolve magic methods is to look for them on the class, not on the instance:

```diff
-hasattr(out, '__fspath__')
+hasattr(type(out), '__fspath__')
```

Well, *even more* Pythonic is to just use `isinstance` with the abstract base class:

```python
isinstance(out, os.PathLike)
```

mentalisttraceur:
Based on the comment in #579 about still supporting Python 2, `isinstance(out, os.PathLike)` is out, so `hasattr(type(out), '__fspath__')` is probably the closest to the Right(tm) test we can do with the same code.

PR #599, opened 2022-02-11
fix: Rename cmd to command in OProc.__repr__

DarkaMaul:
Fix the AttributeError raised when using __repr__ in "OProc".

This issue is annoying when you use a debugger and it tries to represent every object in the frame.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -2309,7 +2309,7 @@ class OProc(object):
             )
 
     def __repr__(self):
-        return "<Process %d %r>" % (self.pid, self.command[:500])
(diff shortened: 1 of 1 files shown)


amoffat:
The tests are failing.

amoffat:
Thanks @DarkaMaul. Can you give me an example/screenshot of the output difference that this makes to help me better understand the problem?

DarkaMaul:
Actually, I think the problem is coming from my IDE and not sh.
I will try to investigate and report later.

PR #601, opened 2022-03-16
Explain type of argument passed to output callback

honnix:
Not sure whether this is already mentioned in other parts of the doc, and I think it could be useful to mention it where we explain callback.


honnix:
@amoffat Wdyt?

amoffat:
Looks good @honnix , thank you

PR #605, opened 2022-05-28
Implement callable support (closes #604)

### LINKED ISSUE #604: Support binding a callable with a baked command
First of all, I'd like to say thanks for this module, it's a lifesaver when working with stuff that only provides command line tools.

Sometimes, I'd like to bind a callable so that I can easily process the output of a command without repetitive boilerplate.

As an example, I'm currently working with the `gcloud` CLI, where I do stuff like this:

```python
iam = sh.gcloud.iam.bake(format='json')
list_of_roles = json.loads(str(iam.roles.list(show_deleted=True, project='my-project')))
```

Having to do stuff like `json.loads(str(...))` for every invocation gets repetitive quickly, and it also makes stuff a little more difficult to reason about.

Therefore, I'd like to have a feature to bind a callable such that `callable(RunningCommand())` is returned. Usage of the API might look like this:

```
iam = sh.gcloud.iam.bake(format='json', _callable=lambda x: json.loads(str(x))
list_of_roles = iam.roles.list(show_deleted=True, project='my-project') # returns a list
```

supriyo-biswas:


amoffatgmi:
According to the related issue, we are not implementing this feature

Issue #304, opened 2016-03-07
sh.Command __unicode__ throws if object isn't completely constructed


virtuald:
We have an [exception logger](http://word.bitly.com/post/69080588278/logging-locals) that captures the inner stack frame and prints out locals. When `CommandNotFound` is thrown in the constructor, the necessary attributes (`_path` and `_partial_baked_args`) haven't been set yet, and so repr(self) throws.

It's a bit of a weird use case, but should be trivial to fix.

amoffat:
Thanks for reporting @virtuald.  [Fix](https://github.com/amoffat/sh/commit/b08586ff8829652d300f908b37ce01f3a8417cbb) is on the `release-1.2` branch and will go out with it

PR #616, opened 2022-08-08
POC: Add lazy resolving of command paths

ecederstrand:
Defer resolving of the actual command path to when the command is called, not when it's imported. This allows providing a custom PATH via _env to customize resolving of the command path.

Remove test case that only tested if the mocker works.

NOTE: this is just a proof of concept of how lazy resolving could work. See discussions in https://github.com/amoffat/sh/pull/602

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -579,22 +579,25 @@ def _which(program, paths=None):
     return found_path
 
 
-def resolve_command_path(program):
-    path = _which(program)
+def resolve_command_path(program, paths=None):
+    path = _which(program, paths=paths)
     if not path:
         # our actual command might have a dash in it, but we can't call
         # that from python (we have to use underscores), so we'll check
         # if a dash version of our underscore command exists and use that
         # if it does
         if "_" in program:
-            path = _which(program.replace("_", "-"))
+            path = _which(program.replace("_", "-"), paths=paths)
         if not path:
             return None
     return path
 
 
 def resolve_command(name, command_cls, baked_args=None):
-    path = resolve_command_path(name)
+    baked_env_paths = None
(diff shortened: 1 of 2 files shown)


sh.py
```diff
@@ -1307,17 +1309,6 @@ def __init__(self, path, search_paths=None):
...
         self.__name__ = str(self)
 
-        if not found:
-            raise CommandNotFound(path)
-
-        # the reason why we set the values early in the constructor, and again
-        # here, is for people who have tools that inspect the stack on
-        # exception.  if CommandNotFound is raised, we need self._path and the
-        # other attributes to be set correctly, so repr() works when they're
-        # inspecting the stack.  issue #304
```
amoffat, line 1317: Does moving this block result in a regression for #304 ?

sh.py
```diff
@@ -3307,13 +3305,50 @@ def __getitem__(self, k):
...
+        return LazyResolver(self, k)
+
+    # Methods that begin with "b_" are implementations of shell built-ins that
+    # people are used to, but which may not have an executable equivalent.
+    @staticmethod
+    def b_which(program, paths=None):
+        return _which(program, paths)
+
+
+class LazyResolver(object):
```
amoffat, line 3317: Would you mind adding a docstring of what this is doing please? :)

amoffat:
I think this might have to go into the 2.0.0 branch instead of develop, since it changes the public behavior (though not the interface). If the 1.* users are relying on exceptions to happen on import, this will be unexpected. If you don't disagree, would you mind changing the destination branch to release/2.0.0 please?

amoffat:
Closing for now, but we can revisit in the future. Good work on the POC

Issue #556, opened 2021-02-11
UnicodeDecodeError in stdout/err thread


r4lv:
Dear amoffat,

First, thank you for such an amazing project!

Lately, I've been using *sh* to generate LaTeX documents with `latexmk`, and *sh* gets stuck with an `UnicodeDecodeError` — I think this is related to `_tee`, and some strange stdout output. I'm using *sh* like this:

``` python
cmd = sh.latexmk("-pdf", blabla, _out=sys.stdout, _err_to_out=True, _tee="out", _bg=True, _bg_exc=True)
try:
    cmd.wait()
except KeyboardInterrupt:
    cmd.kill()
except sh.ErrorReturnCode as e:
    # process e.stdout, ...
```

and it never finishes the `wait()`. This is the error I obtain:

```
Underfull \hbox (badness 10000) in paragraph at lines 2--2
Overfull \hbox (0.9766pt too wide) in paragraph at lines 9--10
Exception in thread STDOUT/ERR thread for pid 5456:
Traceback (most recent call last):
  File "/Users/r4lv/.pyenv/versions/3.9.1/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/Users/r4lv/.pyenv/versions/3.9.1/Python.framework/Versions/3.9/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/r4lv/.pyenv/versions/3.9.1/Python.framework/Versions/3.9/lib/python3.9/site-packages/sh.py", line 1637, in wrap
    fn(*rgs, **kwargs)
  File "/Users/r4lv/.pyenv/versions/3.9.1/Python.framework/Versions/3.9/lib/python3.9/site-packages/sh.py", line 2585, in output_thread
    done = f.read()
  File "/Users/r4lv/.pyenv/versions/3.9.1/Python.framework/Versions/3.9/lib/python3.9/site-packages/sh.py", line 3049, in read
    self.write_chunk(chunk)
  File "/Users/r4lv/.pyenv/versions/3.9.1/Python.framework/Versions/3.9/lib/python3.9/site-packages/sh.py", line 3026, in write_chunk
    self.should_quit = self.process_chunk(chunk)
  File "/Users/r4lv/.pyenv/versions/3.9.1/Python.framework/Versions/3.9/lib/python3.9/site-packages/sh.py", line 2928, in process
    handler.write(encode(chunk))
[...]

PR #628, opened 2022-10-19
Make fd/file_chunk_consumer respect decode_errors

felixonmars:
This fixes the example I provided at
https://github.com/amoffat/sh/issues/556#issuecomment-1281119556 by making fd/file_chunk_consumer respect decode_errors just like the other consumers.


ecederstrand:
Thanks for this! It looks good.

It would be great if we can get some test coverage of this. There's a `test_decode_error_handling` test case that can be added to.

felixonmars:
I have adjusted the test case to correctly cover this.

ecederstrand:
Thanks for the contribution!

PR #644, opened 2023-02-10
add missing files to sdist

mgorny:
Add missing documentation and test files to sdist archives.  The latter are needed by Linux distributions to test the package.


amoffat:
Thanks @mgorny. Can you please bump the micro version number in `pyproject.toml`? Otherwise the deploy CD step will fail.

mgorny:
Thanks! I presume the previous comment should be ignored, correct?

amoffat:
Yes i bumped it. Thanks for your help 🙏

Issue #649, opened 2023-02-11
Performance regression introduced in 2.0.0


ypsah:
Hi,

I am (happily) using `sh` in my [`simple-git-versioning`](https://gitlab.com/ypsah/simple-git-versioning) project, and since version 2.0, I have noticed a significant slowdown in test execution: my 16 tests used to run in 1 second, and they now take 14 seconds.

Using `git bisect`, I tracked the issue down to e9511de5ca9b1d1bf0b42da51fa40073a73fa4d6, and more specifically, this part of the change:

```patch
@@ -2371,7 +2414,7 @@ class OProc(object):
             # thread, which is attempting to call wait(). by introducing a tiny sleep
             # (ugh), this seems to prevent other threads from equally attempting to
             # acquire the lock. TODO find out if this is a general python bug
-            time.sleep(0.00001)
+            time.sleep(0.1)
             if self.exit_code is not None:
                 return False, self.exit_code
             return True, self.exit_code
```

Reverting this one line gets the execution speed back down to 1 second for my 16 tests.

JacobHayes:
I'm seeing this too - I have a job that normally takes ~4m but after the upgrade it ran >8h before I cancelled. I reran after downgrading and it ran in ~4m again. I tried out #650 and it's been stuck for ~20m. I've checked the CPU usage and it stays consistently low (around .03% for python, 0% for the subprocess). Perhaps the `.acquire(timeout=0.1)` is still slow or perhaps there are other regressions.

This process has `_out`, `_err`, and _in` function handlers, maybe those compound sleeps?

amoffatgmi:
Btw do we have any minimum reproducible example we can try our changes against to see the performance differences?

JacobHayes:
Looks like any test with a generator `_in` will do (I didn't test other `_in` variants):
```python
import sh

sh.cat(_in=("a" for _ in range(100)))
```
```sh
$ pip3 install 'sh<2' 2>/dev/null | grep "Successfully installed"
Successfully installed sh-1.14.3
$ time python3 run.py

________________________________________________________
Executed in   58.41 millis    fish           external

$ pip3 install 'sh>2' 2>/dev/null | grep "Successfully installed"
Successfully installed sh-2.0.1
$ time python3 run.py

________________________________________________________
Executed in   10.35 secs      fish           external

[...]

PR #650, opened 2023-02-11
use threading.Lock.acquire's timeout for backpressure rather than sleep

ypsah:
fixes: #649

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -2491,16 +2491,12 @@ class OProc(object):
         # so essentially what we're doing is, using this lock, checking if
         # we're calling .wait(), and if we are, let .wait() get the exit code
         # and handle the status, otherwise let us do it.
-        acquired = self._wait_lock.acquire(False)
+        #
+        # Using a small timeout provides backpressure against code that spams
+        # calls to .is_alive() which may block the main thread from acquiring
+        # the lock otherwise.
+        acquired = self._wait_lock.acquire(timeout=0.1)
         if not acquired:
-            # this sleep here is a hack. occasionally, is_alive can get called so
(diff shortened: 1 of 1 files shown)


sh.py
```diff
@@ -2491,16 +2491,12 @@ def is_alive(self):
         # so essentially what we're doing is, using this lock, checking if
         # we're calling .wait(), and if we are, let .wait() get the exit code
         # and handle the status, otherwise let us do it.
-        acquired = self._wait_lock.acquire(False)
```
amoffat, line 2494: There must have been a reason why `blocking=False` and no timeout, but I can't think of it...

amoffat (approved):

sh.py
```diff
@@ -2491,16 +2491,12 @@ def is_alive(self):
...
+        #
+        # Using a small timeout provides backpressure against code that spams
+        # calls to .is_alive() which may block the main thread from acquiring
+        # the lock otherwise.
+        acquired = self._wait_lock.acquire(timeout=0.1)
```
JacobHayes, line 2498: I still had to set this back down to:
```suggestion
        acquired = self._wait_lock.acquire(timeout= 0.00001)
```
as it was before https://github.com/amoffat/sh/commit/e9511de5ca9b1d1bf0b42da51fa40073a73fa4d6.

Given we're using the `acquire` provided timeout, is that fine now?

JacobHayes (changes requested):

JacobHayes:
Ah, a minute too late 😁

amoffatgmi:
There's still time @JacobHayes, it's only merged into the `develop` branch. It won't ship until it lands in the main branch. If you have a proposed change, make it and tag @ypsah

PR #675, opened 2023-05-12
Make sure new_group never creates new session

### LINKED ISSUE #671: test_new_session_new_group fails on OpenIndiana
The following test consistently fails on OpenIndiana:
```
FAIL: test_new_session_new_group (tests.test.FunctionalTests)
  File "$(BUILDDIR)/tests/test.py", line 2776, in test_new_session_new_group
    self.assertTrue(event.is_set())
AssertionError: False is not true
```
~~It is easily possible that the problem is neither in `sh`, nor its tests, but in OpenIndiana instead.  Unfortunately, I do not know how to prove that.  IOW, I do not understand what exactly is the test trying to do.~~

mtelka:
According to [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/setpgrp.html) it is unspecified whether the `setpgrp()` function should behave like `setpgid(0,0)` or `setsid()`.  While the `sh` project assumes it is always like `setpgid(0,0)` this is not the case for operating systems like illumos or Solaris.

This fixes the `new_group` to newer create new session on POSIX compliant operating systems.

Please note that `setpgrp()` is currently considered obsolete by POSIX.

Fixes #671

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -2037,7 +2037,7 @@ class OProc(object):
                 if new_session:
                     os.setsid()
                 elif new_group:
-                    os.setpgrp()
+                    os.setpgid(0, 0)
 
                 sid = os.getsid(0)
                 pgid = os.getpgid(0)


amoffat:
Thanks for the contribution. At first look, this seems like a safe change. Are you aware of any platform differences that could be negatively impacted by this change?

mtelka:
No.

PR #679, opened 2023-06-05
[doc] Add example for Piping to STDIN migration

honnix:
To give a concrete example for users to easily follow.


honnix:
@ecederstrand Does this make sense? Thank you.

amoffat:
Thanks @honnix , this is helpful :pray:

PR #709, opened 2023-11-16
Fix glob

### LINKED ISSUE #708: glob wrapper breaks glob(.., root_dir=someething)
Current glob wrapper only considers the "recursive" argument, but glob.glob supports much more. When using the sh module, it breaks calls that use other parameters too.

Pistahh:
Fixes #708

glob.glob has many possible arguments but the current wrapping solution doesn't cover all them.

This change directly passes all args (positional and kw) as-is to the wrapped function.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -457,12 +457,12 @@ class GlobResults(list):
         list.__init__(self, results)
 
 
-def glob(path, recursive=False):
-    expanded = GlobResults(path, _old_glob(path, recursive=recursive))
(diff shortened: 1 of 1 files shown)


amoffat:
You might need to suppress the linting on the line that failed, since the glob signature is different over the different python versions

amoffat (approved):

amoffat:
Thanks for the contribution @Pistahh :+1:

PR #718, opened 2024-03-13
drop use python<=3.7 syntax

kloczek:
Despite what is mentioned in pyproject.toml that `sh` supports only python 3.8.x `sh` code still uses older syntax.
Pass all code over `pyupgrade --py38`.

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -106,7 +106,7 @@ def get_num_args(fn):
     return len(inspect.getfullargspec(fn).args)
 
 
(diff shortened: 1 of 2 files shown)


ecederstrand:
Thanks for the contribution!

black tripped over a single line because it wants double quotes: `_unicode_methods = set(dir(''))`

If you push that change, then I'll merge.

kloczek:
Thank you 😄

Issue #754, opened 2025-04-27
FR: complete documentation in a single page for LLMs


thiswillbeyourgithub:
Hi,

I've been interested in sh for a while and it seems really awesome.

Notably I'm interested in using it to make very concise wrapper for cli tools, for example [to wrap edgevpn into a python lib](https://github.com/mudler/edgevpn/issues/525)

The limiting factor for such side projects is how profficient LLMs are with it and so far it seems that it's not quite popular enough to be reliably used by  LLMs (my goto being [aider-chat](https://aider.chat/)).

Hence, to use `sh`, the documentation has to be provided manually to the LLM, so would you be interested in making, in the docs, a single page compilation of everything an LLM might want? 

To give an example, I am the dev behind [wdoc](https://github.com/thiswillbeyourgithub/wdoc/) and because it contains many different use cases I made a [Single Page Documentation](https://wdoc.readthedocs.io/en/latest/all_docs.html#) page in the website.

I'm thinking it could be done with a simple commit to add a page in the website.

Thanks and have a nice day!

amoffat:
Hi there. If you want to write it, and it is high quality documentation, I will merge it.

thiswillbeyourgithub:
Great! The thing is, there's not much to write. It's just a matter of adding a page to the docs that contain other pages. [See this for example](https://github.com/thiswillbeyourgithub/wdoc/blob/main/docs/source/all_docs.rst). It's kinda awkward to write that without seeing the readthedocs website update imho but if you want me to i'll try.

amoffat:
I don't think I understand. If it's just a page linking to other pages, why can't the LLM system use the existing docs, which are links to other pages?

Also you can clone the repo and build the docs locally for interactive feedback.

thiswillbeyourgithub:
I'll try to be clearer:

- this [all_docs.rst](https://github.com/thiswillbeyourgithub/wdoc/blob/main/docs/source/all_docs.rst) file includes the pages readme, examples, help, parse_file_help and api.
- The rendered webpage can be seen [here](https://wdoc.readthedocs.io/en/latest/all_docs.html#)
- As you can see, the effort required to create this page is minimal: it just concatenates the already written documentation. It does not `link` them, it contains them.
- The reason I'm making this FR is to ask to make it as simple as copy+pasting a url to get all the documentation of `sh`. Yes it is possible currently to paste each page, or to use things like [repomix](https://repomix.com/) to convert repositories into single text but this adds friction.
[...]

PR #755, opened 2025-04-29
Full documentation as a single page for LLMs

thiswillbeyourgithub:
- **whitespace**
- **doc: add full doc as a single page for LLMs**

As discussed in #754

On my local build, done with `cd docs` then `make clean html`, there's the header issue but it might be just me, might be a readthedocs bug and might happen in the website version, and is not that bad anyway IMHO.

edit: put in draft again until I double check the errors when building the docs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,6 +36,8 @@ systems - Linux, macOS, BSDs etc. Specifically, Windows is not supported.
 
 `Complete documentation here <https://sh.readthedocs.io/>`_
 
+`Full documentation on a single page for LLM-assisted coding here <https://sh.readthedocs.io/en/latest/fulldoc.html>`_
+
 Installation
 ============
 
--- a/docs/source/fulldoc.rst
+++ b/docs/source/fulldoc.rst
@@ -1,0 +1,35 @@
+Full Documentation
+==================
+
+This single page repeats the full documentation for `sh <https://github.com/amoffat/sh/>`, making it easier to put into an LLM's context window. There is nothing on this page that is not mentionned already elsewhere on this site, it's just reorganized as a single page.
+
+.. include:: index.rst
+
+.. include:: tutorials/interacting_with_processes.rst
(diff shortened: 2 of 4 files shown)


thiswillbeyourgithub:
Okay so the current pr is fine. I think it does not interfere at all with the current doc, just adds the new page and reference to it. But apparently the actual good way of doing this would be to use a [custom readthedocs builder](https://docs.readthedocs.com/platform/stable/config-file/v2.html) and that might be quite involved. And would be quite a bit harder for me because I don't have access to readthedocs.

amoffat (approved):

thiswillbeyourgithub:
Thanks for being open to the idea! I tried clicking on the link in the readme and it 404, has the doc not automatically rebuilt itself?

amoffat:
The merge broke the docs. It was the only change in the past 2 months, since the docs last built successfully

image

thiswillbeyourgithub:
I noticed that in my wdoc I have the theme in my requirements for the doc: https://github.com/thiswillbeyourgithub/wdoc/blob/main/docs/requirements.txt

Maybe this would help?

Are you sure the build was succesful previously because I thought I saw differences between the local build I have made even before started coding and what I was seeing on the online version.

thiswillbeyourgithub:
Also iirc my local build was using sphinx 8.2.3

amoffat:
You're right, I misread.. looks like the docs have been broken for awhile. Ay....

thiswillbeyourgithub:
Try maybe with those in the req:
Sphinx                        8.2.3
sphinx-rtd-theme              3.0.2

thiswillbeyourgithub:
BTW, though not very used to readthedocs I am willing to help you out of you give me the error messages etc. This is usually something dumb.

amoffat:
This has shipped, thank you for your patience https://sh.readthedocs.io/en/latest/fulldoc.html

PR #756, opened 2025-05-08
Quote arguments with whitespace when printing command

peterdemin:
Currently, printing baked commands doesn't represent the way arguments are actually passed down to the program when values contain spaces.

The solution is to take such values in single quotes.

This is primarily to be able to copy printed commands and run them in a shell.

BEFORE:

```python
>>> from sh import ls
>>> print(ls.bake('How I Met Your Mother'))
/bin/ls How I Met Your Mother
```

AFTER:

```python
>>> print(ls.bake('How I Met Your Mother'))
/bin/ls 'How I Met Your Mother'
```

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -1391,10 +1391,10 @@ class Command:
         return fn
 
     def __str__(self):
-        baked_args = " ".join(self._partial_baked_args)
-        if baked_args:
-            baked_args = " " + baked_args
-        return self._path + baked_args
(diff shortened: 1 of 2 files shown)


amoffat:
Cool, thanks for the contribution. It may take a little while for it to land in the main branch, as I have some maintenance tasks to do before then.

amoffat (approved):

PR #761, opened 2025-06-22
Add 'x' command argument for tracing

ashermancinelli:
Similar to `set -x` in bash, echos the command to stdout before executing anything.

One of my biggest complaints about using sh in place of shell scripting is that I can't bake-in something that echos the commands that are about to be run, so my scripts are littered with prints. I would love to be able to add `_x=True` to my bake arguments and see what's going to be run.

Given that `set -x` is specific to the shells I'm used to, I could also imagine `_trace=True` or `_explicit=True` being more approachable. Thanks in advance for any suggestions!

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -1273,6 +1273,8 @@ class Command:
         # sometimes we may return just a plain unicode string
         "return_cmd": False,
         "async": False,
+        # whether to echo each command before executing it
+        "x": False,
     }
 
     # this is a collection of validators to make sure the special kwargs make
@@ -1478,6 +1480,8 @@ class Command:
         final_args = split_args
 
         cmd.extend(final_args)
+        if call_args["x"]:
+            print(' '.join(cmd))
 
         # if we're running in foreground mode, we need to completely bypass
         # launching a RunningCommand and OProc and just do a spawn
--- a/sh.py
+++ b/sh.py
@@ -1481,7 +1481,7 @@ class Command:
 
         cmd.extend(final_args)
         if call_args["x"]:
-            print(' '.join(cmd))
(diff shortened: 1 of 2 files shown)


ashermancinelli:
<details><summary>Unrelated FWIW:</summary>


adding this section to pyproject.toml would make it easier to develop this with uv
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 0e8bb2f..eac408e 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[project]
+name = "sh"
+version = "2.2.2"
+
 [tool.poetry]
 name = "sh"
 version = "2.2.2"
```

uv complains with
```
> uv sync                          
error: No `project` table found in: `~/workspace/sh/pyproject.toml`
```

</details>

ecederstrand:
@amoffat may have a different opinion, but in my opinion `sh` is a subprocess replacement, not a bash replacement. As such, implementing `-x` isn't entirely in line with the purpose of `sh`.

If you want to print commands before execution, you can enable logging: https://sh.readthedocs.io/en/latest/sections/faq.html#how-do-i-see-the-commands-sh-is-running

ashermancinelli:
That works for me, thank you!

ashermancinelli:
If someone comes across this in the future, [this gist might be useful.](https://gist.github.com/ashermancinelli/d45496a4e73f0b1dc747d7e71a553821)

```python
with SetX: # prints commands from 'sh' module like 'set -x' does
    sh.opt('foo.ll', '-S', '-o', '-', passes='instcombine')
```

PR #770, opened 2026-02-22
Fix boolean bake override not working

### LINKED ISSUE #753: Cannot override boolean baked arguments
According to `.bake()` docs

> special keyword arguments can be temporary baked and additionally be overridden in __call__ or in subsequent bakes (basically setting defaults)

However, when using boolean keyword arguments this happens:
```python
>>> from sh import ls
>>> ls(a=True) # for reference
'.  ..  Makefile  requirements.txt  source\n'
>>> ls.bake(a=False)(a=True) # works as expected, later one prevails
'.  ..  Makefile  requirements.txt  source\n'
>>> ls(a=False) # for reference
'Makefile  requirements.txt  source\n'
>>> ls.bake(a=True)(a=False) # True prevails, but should have been overriden
'.  ..  Makefile  requirements.txt  source\n'
```

bysiber:
When a boolean argument is baked as `True` (e.g. `ls.bake(a=True)`), calling with `a=False` should override it and drop the flag. Currently the `False` value is silently discarded by `compile_args` while the baked `-a` flag remains in the argument list, so the override has no effect.

This fixes it by filtering out previously baked flags when a `False` override is passed, both in `bake()` (for chained bakes) and `__call__()` (for call-time overrides).

Fixes #753

### DIFF
--- a/sh.py
+++ b/sh.py
@@ -1388,6 +1388,19 @@ class Command:
         sep = call_args.get("long_sep", self._call_args["long_sep"])
         prefix = call_args.get("long_prefix", self._call_args["long_prefix"])
         fn._partial_baked_args.extend(compile_args(args, kwargs, sep, prefix))
+
+        # filter out previously baked args that are overridden with False
+        for key, val in kwargs.items():
+            if val is not False:
+                continue
+            if len(key) == 1:
+                flag = "-" + key
+            else:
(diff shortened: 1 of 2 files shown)


amoffat:
This has landed in 2.2.5. Thanks for highlighting the issue

Issue #778, opened 2026-06-08
2.2.5 release lost tests from source distribution


mgorny:
The 2.2.5 release has lost all the additional files from the source distribution:

```diff
--- /tmp/old	2026-06-08 05:45:26.397507522 +0200
+++ /tmp/new	2026-06-08 05:45:30.459327093 +0200
@@ -1,14 +1,8 @@
-CHANGELOG.md
-images/logo-230.png
-images/logo-big.png
+.gitignore
 LICENSE.txt
-Makefile
-MIGRATION.md
 PKG-INFO
 pyproject.toml
+py.typed
 README.rst
 sh.py
-tests/Dockerfile
-tests/__init__.py
-tests/sh_test.py
-tox.ini
+sh.pyi
```

Most notably, it lost [tests that are required for downstream packaging](https://packaging.python.org/en/latest/discussions/downstream-packaging/#provide-complete-source-distributions).

amoffat:
Thank you for reporting. I moved from poetry to uv, and it's clear the build config didn't carry over

mgorny:
I think you meant `hatchling`? FWICS you need to add the files to `[tool.hatch.build.targets.sdist]`.

PR #779, opened 2026-06-08
include missing sdist files

amoffat:
https://github.com/amoffat/sh/issues/778

@mgorny How does this look?


mgorny:
Thanks. The 2.2.6 release looks good.

PR #783, opened 2026-07-16
Drop support for Python <3.10 (EOL)

jorenham:
Hi there,

My initial plan was to open a PR with some static typing improvements. But then I noticed that you were still supporting two EOL Python versions: 3.9 (EOL since 2015-10) and 3.8 (EOL since  2024-10). Having to support 3.8 makes typing a lot more difficult and verbose, so I figured I'd open this PR first. 

https://devguide.python.org/versions/

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
     "ghcr.io/devcontainers/features/copilot-cli:1": {},
     "ghcr.io/devcontainers/features/github-cli:1": {}
   },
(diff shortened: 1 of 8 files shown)


amoffat (approved):

amoffat:
good call!

amoffat:
Also I'm excited in what you cook up with typing improvements. I know they can be better :)

jorenham:
Thanks for the quick review! I'll look into the typing improvements next week

PR #785, opened 2026-07-20
Modernize type annotations in the stubs

jorenham:
Following #783, this replaces deprecated `typing` special forms with modern [PEP 604](https://peps.python.org/pep-0604/) type annotations, and imports the collection types from `collections.abc` instead of from `typing` (deprecated aliases).

Don't take this the wrong way, but the stubs looks very much AI generated. But the thing is: AI is absolutely horrible when it comes to typing in Python in my experience. This is a good example of that actually, because there are quite a lot of issues. For example, when I run `stubtest sh` (stubtest is part of mypy), it reports 357 errors. There's also a lot of unnecessary code duplication that could easily be resolved by introducing a couple of type aliases. So after this, I'd really like to start improving this stubs (I can't wait TBH haha).

### DIFF
--- a/src/sh/__init__.pyi
+++ b/src/sh/__init__.pyi
@@ -9,28 +9,12 @@ patterns like ``from sh import ls`` to type-check cleanly.
 """
 
 import threading
+from collections.abc import AsyncIterator, Callable, Generator, Iterable, Set
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


amoffat:
cool, i look forward to seeing the improvements :+1:

PR #786, opened 2026-07-20
Dedupe type annotations in stubs using type aliases

jorenham:
As promised in #785, this deduplicates several beefy annotations that were repeated 10+ times by introducing [PEP 613](https://peps.python.org/pep-0613/) type aliases for them.

I'm not sure about the type alias names though :thinking:


amoffat:
lgtm

PR #787, opened 2026-07-21
Fix metaclass issues in the type stubs

jorenham:
Continuing the typing pilgrimage that started at #785, this takes care of the largest class of stubtest errors, bringing the total of 357 errors down to 34.

Before:

```console
$ stubtest sh
<errors omitted for sanity>
Found 357 errors (checked 1 module)
```

After:

```console
$ stubtest sh
<errors omitted for sanity>
Found 34 errors (checked 1 module)
```

:mage:


amoffat:
What does the metaclass enable/fix in a stub file?

jorenham:
> What does the metaclass enable/fix in a stub file?

It's mostly a correctness thing; stubs should accurately describe the runtime situation, including metaclasses. 

It might make a difference in type-safety in case of `__slots__`, but I doubt that'll be relevant here. 

So TBH, it's mostly to appease stubtest. The alternative would be to manually add these to the stubtest "allowlist" (a list of ignored errors), which sounds like more work than simply adding these metaclasses :shrug:

PR #788, opened 2026-07-22
Fix `Command` stubtest errors

jorenham:
Following #787, here's the next batch of stubtest fixes. This time for the `sh.Command` class.

before:

```console
$ stubtest sh
<errors omitted for sanity>
Found 34 errors (checked 1 module)
```

after:

```console
$ stubtest sh
<errors omitted for sanity>
Found 25 errors (checked 1 module)
```

There's one place where I deviated from runtime, which is for the return type of `__class_getitem__`. At runtime it directly returns `cls`, but it's better to have it return a `types.GenericAlias` so that e.g. `print(Command[str])` will be printed as `Command[str]`. Currently it also allows nonsensical expressions like `Command[str][str][str]`, which won't be possible anymore if we'dd change that to return a `GenericAlias`.
Do you want me to take care of that as well? And if so, should I put it in here, or in a separate PR?


amoffat:
Sure you can put it in this PR if it's easier. Also, given that I had never heard about `stubtest` before, is it reasonable to have another lint CI step for it?

jorenham:
> given that I had never heard about stubtest before,

FWIW; It's used by [typeshed](https://github.com/python/typeshed) itself, and we also use it at NumPy, for example, so it's pretty decent stuff.

> is it reasonable to have another lint CI step for it?

Yea that's indeed what I was planning on after fixing the errors :)

PR #789, opened 2026-07-22
Remove Python 2 dunder methods

jorenham:
`__nonzero__` and `__long__` are not a thing anymore :)

BTW, removing the `__bool__` methods entirely is also an option, because then it'll just fallback to checking `__len__` for truthiness checks.


amoffat:
sh is showing its age :smile:

jorenham:
> sh is showing its age 😄

And that's awesome. There are way too many abandoned open source projects. So thanks for keeping sh alive for this long (punintended :P).

PR #791, opened 2026-07-23
Fix remaining stubtest errors

jorenham:
As promised in #790, here's the last PR in the stubtest series, bringing the stubtest errors down to...

(drumroll please)

... **0** :tada:!!

```console
$ stubtest sh
Success: no issues found in 1 module
```

Oh, one thing I wasn't really sure about is the removal of `SignalException_SIGUSR1` and `SignalException_SIGUSR2`. Stubtest wanted me to, so I simply did. But I was left wondering whether maybe it was a system-dependent thing that's should stay?

Anyway, aftyer this, I plan on working on adding stubtest to the CI pipeline so we can ensure that the stubs stay in sync with the runtime API.

### DIFF
--- a/src/sh/__init__.pyi
+++ b/src/sh/__init__.pyi
@@ -13,8 +13,20 @@ import threading
 from collections.abc import AsyncIterator, Callable, Generator, Iterable, Set
 from contextlib import contextmanager
 from queue import Queue
-from types import GenericAlias, TracebackType
-from typing import IO, Any, ClassVar, Final, Generic, Literal, Self, TypeAlias, overload
+from types import GenericAlias, ModuleType, TracebackType
+from typing import (
+    IO,
+    Any,
+    ClassVar,
+    Final,
+    Generic,
+    Literal,
+    Protocol,
+    Self,
+    TypeAlias,
+    overload,
+    type_check_only,
+)
 
 from typing_extensions import TypeVar
 
@@ -31,6 +43,7 @@ _CommandArgPreprocess: TypeAlias = Callable[..., tuple[list[Any], dict[str, Any]
 # Version / metadata
 # ---------------------------------------------------------------------------
 
+__all__ = []
 __version__: str
 __project_url__: str
 DEFAULT_ENCODING: str
@@ -42,6 +55,7 @@ DEFAULT_ENCODING: str
 class ForkException(Exception):
     def __init__(self, orig_exc: str) -> None: ...
 
+@type_check_only
 class ErrorReturnCodeMeta(type): ...
 
 class ErrorReturnCode(Exception):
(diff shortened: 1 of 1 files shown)


amoffat:
Awesome!! 🎉🎉🎉

We should leave SIGUSR1 and SIGUSR2. It's strange that stubtest was complaining about those, since they should be like the other signals. Otherwise this looks great :raised_hands:

jorenham:
Hmm, I think that's a runtime bug then:

```pycon
>>> import sh
>>> sh.SignalException_SIGKILL
<class 'sh.SignalException_SIGKILL'>
>>> sh.SignalException_SIGUSR1
```

```pytb
  File "/home/joren/Workspace/sh/src/sh/__init__.py", line 390, in get_exc_from_name
    return rc_exc_cache[name]
           ~~~~~~~~~~~~^^^^^^
KeyError: 'SignalException_SIGUSR1'

During handling of the above exception, another exception occurred:

  File "/home/joren/Workspace/sh/src/sh/__init__.py", line 399, in get_exc_from_name
    rc = -int(rc_or_sig_name)
          ~~~^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'SIGUSR'

During handling of the above exception, another exception occurred:

    sh.SignalException_SIGUSR1
  File "/home/joren/Workspace/sh/src/sh/__init__.py", line 3742, in __getattr__
    return self.__env[name]
           ~~~~~~~~~~^^^^^^
  File "/home/joren/Workspace/sh/src/sh/__init__.py", line 3487, in __getitem__
    exc = get_exc_from_name(k)
  File "/home/joren/Workspace/sh/src/sh/__init__.py", line 401, in get_exc_from_name
    rc = -getattr(signal, rc_or_sig_name)
          ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'signal' has no attribute 'SIGUSR'. Did you mean: 'SIGUSR1'?
```

This seems to be caused by the regex at line 369:
 https://github.com/amoffat/sh/blob/46107ba881d043a241e48551a020954846e0c996/src/sh/__init__.py#L369

The `SIG[a-zA-Z]+` branch has no digits, and since it's `re.match` (prefix match, not fullmatch), `"SignalException_SIGUSR1"` matches with the 2nd group, i.e. `"SIGUSR"`, dropping the trailing digit. But `signal.SIGUSR` isn't a thing, so that leads to the `AttributeError` above.

amoffat:
Ok fixed, you should be good to proceed

amoffat (approved):

jorenham:
Thanks, I really appreciate your quick responses :)

PR #794, opened 2026-07-23
Run `stubtest` and `pyrefly coverage check` in CI (and tox)

jorenham:
As promised in #791:

> I plan on working on adding stubtest to the CI pipeline so we can ensure that the stubs stay in sync with the runtime API.

... and so it shall be done.

Oh and while we're at it, it might also be a good idea to ensure 100% type coverage (i.e. ensure that everything in the public API is annotated). For that we could use `pyrefly coverage check`  ([docs](https://pyrefly.org/en/docs/report/); but don't mind the "experimental" status, which doesn't really apply anymore, ~and I think I'll go ahead and open a PR to remove that right now~ and I opened a PR for that: https://github.com/facebook/pyrefly/pull/4269):

```console
$ uvx pyrefly coverage check src/sh
 INFO type coverage 100.00% (178 of 178 typable)
```

When I remove a random annotation in the stubs, here's what you'll see:

```console
$ uv run stubtest sh
Success: no issues found in 1 module

$ uvx pyrefly coverage check src/sh
 WARN `sh.glob` is not fully typed [coverage-partial]
    --> src/sh/__init__.pyi:1393:1
     |
1393 | def glob(path, *args: Any, **kwargs: Any) -> list[str]: ...
     | -----------------------------------------------------------
     |
ERROR type coverage 99.44% (177 of 178 typable) is below the 100.00% threshold
```

And FWIW; we're also using it in numpy and scipy-stubs, and it has been very helpful, and was able to catch things that stubtest didn't catch.

So if you want I could also add that as well?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,7 @@ jobs:
           uv run ruff format --check src/sh tests/
           uv run mypy src/sh/
           uv run stubtest sh
+          uv run pyrefly coverage check src/sh
           uv run rstcheck README.rst
 
   test:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,6 +30,7 @@ jobs:
           uv run ruff check src/sh/ tests/
           uv run ruff format --check src/sh tests/
           uv run mypy src/sh/
+          uv run stubtest sh
           uv run rstcheck README.rst
 
   test:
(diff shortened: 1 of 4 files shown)


jorenham:
CI failure seems unrelated:

```
coveralls.exception.CoverallsException: Could not submit coverage: 500 Server Error: Internal Server Error for url: https://coveralls.io/api/v1/jobs
```

amoffat:
> So if you want I could also add that as well?

Yeah, let's do it :+1:  Thank you! This is awesome. People have had gripes about sh typing in modern IDEs for a long time, and your work here really buttons it up.

amoffat (approved):

PR #797, opened 2026-09-12
Fix repeated boolean overrides in baked commands

oodadoudou:
Follow-up to #770: after baking a flag as `False` and then `True`, a later `False` leaves it enabled. The index keeps the first `False` and misses later overrides. Track the last `False` so both `bake()` and call-time overrides remove earlier `True` flags.

The regression checks the child process's arguments for short and long flags, including disabling, re-enabling, and reusing the original baked command.

### DIFF
--- a/src/sh/__init__.py
+++ b/src/sh/__init__.py
@@ -1417,21 +1417,21 @@ class Command:
         if extra_args or extra_kwargs:
             all_layers.append((extra_args, extra_kwargs))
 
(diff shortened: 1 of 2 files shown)


amoffat:
This was on my backlog to look into. Is this a reason this was closed @oodadoudou ?

oodadoudou:
Yes, that is why the PR was closed. I am not planning to pursue it further, so no review is needed. Thanks for checking.

